PR #26, opened 2011-05-26
Now it is possible to filter attribute values by passing callable in BLEACH_ATTRIBUTES

olomix:
Example: 

BLEACH_ATTRIBUTES = {
    'a': ['href', 'title', 'name', 'target'],
    'abbr': ['title'],
    'acronym': ['title'],
    'img': ['src', 'alt'],
    'iframe': lambda name, val: name=='src' and re.match(r'http://www.youtube.com/', val),
    }

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -42,7 +42,9 @@ to customize their behavior.
 | ``attributes``     | A whitelist of HTML attributes. Either a list, in     |
 |                    | which case all attributes are allowed on all elements,|
 |                    | or a dict, with tag names as keys and lists of allowed|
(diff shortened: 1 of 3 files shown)


cpury:
I was looking for a functionality to only allow certain iframes in user input (i.e. youtube embeds). The fix proposed by olomix is the best I found so far. However, the iframe object is still kept in the output (without the src attribute) if the src is not youtube. I allow width and height attributes, so this can be pretty ugly...

Is there a way to completely filter iframes that are from a non-whitelisted src?

PR #64, opened 2012-04-30
Add sanitize option to linkify()

ThiefMaster:
Sometimes sanitization is not wanted when linkifying. Add an option to enable it only if necessary.
It defaults to off since that makes more sense in my optinion (you do not expect a method `linkify` to escape some tags!)


ganeshaditya1:
Hi,

I ran into the same issue mentioned here while linkifying text containing an iframe(from youtube embed tag). I was wondering why the branch was deleted! 

Thanks.

PR #122, opened 2014-03-25
Allow protocols and svg properties to be configurable.

davestgermain:
I wanted to allow `data` URIs in images, and in order to do that I had to be able to set `allowed_protocols` and plumb it through to the sanitizer. While I was in there, I made the `allowed_svg_properties` attribute configurable, too.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -26,3 +26,4 @@ Patches:
 - Marc DM
 - mdxs
 - Marc Abramowitz
+- Dave St.Germain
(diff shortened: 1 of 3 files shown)


willkg:
Adding support for protocols is being covered more comprehensively in #149. I think I'm inclined to continue moving on that there.

What's the use case for adding svg properties?

willkg:
@davestgermain What's the use case for adding svg properties?

willkg:
I'm going to close this out because it has merge conflicts, unanswered questions and it fails CI.

If you want to continue working on this, can you write up an issue so we can figure out what the problem being solved is, then we can work on solutions.

Thank you! Sorry for not moving this forward faster.

PR #123, opened 2014-04-28
Allow Data URI Schemes

blag:
Allow the 'data' pseudo-protocol as established with https://en.wikipedia.org/wiki/Data_URI_scheme#Format. There is a current PR for html5lib to add 'data' to acceptable_protocols ([PR #148](https://github.com/html5lib/html5lib-python/pull/148)), but this prevents Bleach from incorrectly deleting URI attributes with the 'data' protocol until html5lib is fixed.


willkg:
#122 and #149 both cover changing allowed protocols. We should continue this conversation in one of those two places depending.

Issue #109, opened 2013-10-03
Wishlist: Add the ability to tell if something needed to be "cleaned" or not.


jvanasco:
It doesn't seem possible to test whether or not text needed to be "cleaned"

As part of the tokenization process, `bleach` returns text with the attributes alphabetized, so a string comparison is not possible.

For example:

```
>>> input = """<a href="http://example.com" title="Example Link" class="foo">Example</a>"""
>>> output = bleach.clean(input)
>>> print output
u"""<a class="foo" href="http://example.com" title="Example Link">Example</a>"""
```

so , obviously...

```
>>> input == output
False
```

This makes it impossible to track metrics regarding unsafe content submission -- or alert users if their content is unsafe.

willkg:
If someone wants to tackle this, I suggest doing it in another library and using bleach if that helps and depending on how that effort turns out, maybe we can merge the two at some future date.

Given the above comments, I think we're at a point where it's clear this is impossible to do effectively and thus it's not something I want to add to this library.

PR #131, opened 2014-08-27
Whitelist allowed tags by callable

kampfschlaefer:
This adds the possibility to filter the tags by function, similar to the way its done for attributes.

Our use-case here is that we want to dis-allow `script` and other evil tags but don't want to allow an ever-growing whitelist. And our users often add things like `<username here>` in their texts...

I don't know if there is a better way to tell python that the allowed_tags-function is a static function.

Oh, I also renamed the tests for attributes to `bleach.clean` and added a test for the attribute-filter function.

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -34,7 +34,8 @@ class BleachSanitizerMixin(HTMLSanitizerMixin):
 
         if token['type'] in (tokenTypes['StartTag'], tokenTypes['EndTag'],
                              tokenTypes['EmptyTag']):
-            if (self.allowed_elements(token['name'])
+            if (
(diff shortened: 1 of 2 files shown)


marfire:
Although this issue is long dead, I'll just mention that this ability (using a callable for tags) is essential for allowing the two-pass approach to validation mention in #109. By using a callable that always returns `True` for both tags and attributes, you can provoke bleach to produce unfiltered output to compare to the cleaned output.

Without this, you have to think of and list every possible tag that might exist, now and in the future. That approach is too brittle.

I understand that validation is not a goal of this project, but I'll just mention here for others who are interested that with a few lines of code (including the change in this PR) you can turn bleach into an effective validator.

Issue #129, opened 2014-06-10
feature request - configurable options for HTMLSerializer


jvanasco:
I ran into a weird issue.  It would seem that running something through the bleach multiple times will result in different output.  here's a testcase - https://gist.github.com/jvanasco/2cda35740de8afc0a003

what happens is that the underlying HTMLSerializer will build the serialize the text differently, depending on the output

pay attention to the order of the attributes here:

    <button class="navbar-toggle collapsed" data-target=".navbar-collapse" data-toggle="collapse" type="button">

and here:

    <button data-target=".navbar-collapse" class="navbar-toggle collapsed" data-toggle="collapse" type="button">

html5lib seems to order the attributes "not randomly" based on their initial ordering.  they re-order endlessly between the two .

in our case, we "bleach" user content on input.  

on certain screens, if something had to be bleached, we present the bleached result to the user to approve.  because of this behavior, bleaching the content a second time will result in a document that doesn't match the input document.  

there are workarounds that we employ for now, but it's messy.  we just patched bleach to pass `alphabetical_attributes=True` into HTMLSerializer

it would be great if bleach either had some more default options for HTMLSerialize ( in `_serialize` ) or allowed for some configuration of defaults.

willkg:
The initial issue here is also covered in issue #121. I'm going to close this out since that issue is fixed now.

PR #135, opened 2014-10-31
Add extra_serializer_options to the clean and linkify params

jespino:


willkg:
@jespino I know it's been *ages* since you worked on this, but is it possible for you to fix the merge conflicts? If not, I can take this over and finish it up.

jespino:
Hi, I can take a look to it, thanks

jespino:
Now is up to date, but with the initial approach of extra argument to functions (I understand is the less intrusive option to not break backward compatibility). Thanks for take the time to review it :smile:

jespino:
Any news about it?

willkg:
I screwed up. I didn't spend enough time looking into the context here before asking you to fix merge conflicts.

I just went through everything involved. The issue in #129 is specifically "attribute ordering isn't stable between bleach calls" and generalized into the solution "if only we could pass in additional serializer arguments so we could fix that problem". @jsocol in an earlier comment was pretty concerned about how this solution opens up a huge leaky hole in the API that creates a bunch of problems for us. For a contrived example, if bleach ever were to switch HTML libs, this API change could cause a lot of heartache for users.

Anyhow, the attribute ordering problem is also covered in #121 which was fixed in https://github.com/mozilla/bleach/commit/2df578fdb7d3428cbcbfa578805a2173435d4cca in December 2014. That's available now.

Thus the specific problem that is the crux of #129 seems like a duplicate of #121 and that's fixed so I spent some time looking for other compelling use cases that would require some kind of way to set additional serializer arguments. I don't see evidence of others presently.

Given that, I think I want to push off this kind of change until we have a compelling use case.

Thank you for writing up the issue and working on it! I apologize for screwing up earlier.

Closing this out.

Issue #124, opened 2014-04-28
linkify takes forever with parse_email=True


gaearon:
I presume it's a bug:

    >>> bleach.linkify("<br>test@test.com", parse_email=True)

takes forever to execute on 1.4.

vivazzi:
I have this infinite loop too. Bleach version - 1.4.1.
my example:
```>>> bleach.linkify('<p>test<br/>, email test@mail.ru</p>')```

In debug I have seen next result like this:
```u'" ><a href="<a href="mailto:test@mail.ru" >mailto:test@mail.ru</a>" >test@mail.ru</a>'```
It likes this recursion.

Alhought, next test is passed:
```
>>> bleach.linkify('<p>test email test@mail.ru</p>')
u'<p>test email <a href="mailto:test@mail.ru">test@mail.ru</a></p>'
```

PR #144, opened 2015-01-20
Fix for #124

TintypeMolly:
There is a bug with email parsing #124.
This happens when html element which is direct parent of email string has child tag inside.
Example:
```
<p><span>foo</span>example@gmail.com</p>
<p>qwer<br />example@gmail.com</p>
<br />example@gmail.com
```
Those 3 above hangs when calling `linkfiy` with `parse_email=True`
It is because `link_repl` is called after `email_repl`, making part of email address(`gmail.com`) to an anchor tag again.

At line 236, in the middle of lines that process `node.text`, that kind of unwanted recursion was taken care of, but for `node.tail`, it wasn't.
So I added a `continue` in the middle of lines that process `node.tail`.

Now it doesn't hang, perfectly showing beautiful links.

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -256,6 +256,7 @@ def linkify(text, callbacks=DEFAULT_CALLBACKS, skip_pre=False,
                         # Insert the new nodes made from my tail into
                         # the tree right after me. current_child+1
                         children += adj
+                        continue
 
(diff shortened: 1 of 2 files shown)


jsocol:
Thank you! I squashed the commits and merged them here: 4c7e6325a13737138c682ae8251dcace7a58f9df

PR #159, opened 2015-08-05
Add TLDs from IANA TLD list

Perlmint:
[All valid TLD list](http://data.iana.org/TLD/tlds-alpha-by-domain.txt) provided by ICANN(IANA)  
List was generated by this script : https://gist.github.com/Perlmint/b125232975e8c6ba5243


jsocol:
For a couple of reasons I'd really rather look at using the [public suffix list](https://publicsuffix.org/list/public_suffix_list.dat). One is that the domains aren't IDNA-encoded—at the very least, we need to add the decoded versions, e.g. `বাংলা`, as well as `'xn--54b7fta0cc`.

Perlmint:
I added IDNA-decoded TLDs.

willkg:
@jsocol Any thoughts on where this is at? What's left to do here?

jsocol:
Up to you. I'd still rather make it rely on the public suffix list data—e.g. include the file, read/parse it—so it's in an externally-maintained source that's easier to periodically update, and so it's more accurate (using suffixes instead of pure TLDs). Maintaining the list of encoded and not-encoded TLDs is a pain, keeping it all in `__init__.py` is a pain. But this obviously works, in the short term.

Perlmint:
I changed to use regular expression generated with public suffix list. also include `suffix_update.py`.
Is it looks good?

willkg:
I haven't had a chance to look at this, yet. I'll try to look at it before the end of the year.

suffix_update.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # find idna decodable suffixs
+    idna_encoded_suffix_list = []
+    for suffix in suffix_list:
+        try:
+            encoded = suffix.encode('idna').decode('ascii')
+            if suffix != encoded:
+                idna_encoded_suffix_list.append(encoded)
+        except Exception:
+            pass
```
willkg: What cases will this fail? This catches all exceptions silently and then continues onwards as if nothing happened. If that's the intended behavior, this should at least have a comment explaining what typically gets thrown and why it's ok to ignore that.

willkg: Seems like you don't want to catch all the thing, but rather want to catch `UnicodeEncodeError`. Does that sound right?

suffix_update.py
```diff
@@ -0,0 +1,132 @@
...
+    suffix_list.sort()
+
+    return suffix_list
+
+
+def write_suffix_py(o, suffix_list):
+    '''
+    Write suffix list as python source.
+    '''
+    o.write("# -*- coding: utf-8 -*-\n"
```
willkg: I'm not a big fan of one-letter variables like `o`. Can you rename this variable to something more helpful?

suffix_update.py
```diff
@@ -0,0 +1,132 @@
...
+    # write public suffixes Regular Expression.
+    suffix_list = list(map(lambda x: x.replace('.', '\\.')
+                           .replace('*', '[^.]+').replace('!', ''),
+                           suffix_list))
+    suffix_list.sort()
+    suffix_list.reverse()
+    write('SUFFIXES_RE = (', suffix_list, ')', 3,
+          lambda x, f: ('"%s"' if f else '"|%s"') % '|'.join(x),
+          lambda x: (x, 1))
+    o.write('\n')
```
willkg: Seems like this function is doing more than "write suffix list as python source", but it's pretty complicated and very little of it is commented, so I don't really understand what it's doing and why.

Can you add more explanatory comments as well as a more detailed docstring covering what the output should look like?

suffix_update.py
```diff
@@ -0,0 +1,132 @@
+import sys
```
willkg: This file should include a top-level docstring explaining what it is, what it does and how to use it.

suffix_update.py
```diff
@@ -0,0 +1,132 @@
...
+                           suffix_list))
+    suffix_list.sort()
+    suffix_list.reverse()
+    write('SUFFIXES_RE = (', suffix_list, ')', 3,
+          lambda x, f: ('"%s"' if f else '"|%s"') % '|'.join(x),
+          lambda x: (x, 1))
+    o.write('\n')
+
+if __name__ == "__main__":
+    from argparse import ArgumentParser
```
willkg: Python 2.6 doesn't have argparse so this won't run in that environment. I think that's probably ok, but a note should be added to the top-level docstring for this file.

willkg:
Have you thought about testing this? How can we verify that the output is good in an automated way?

willkg:
I think that's everything I can think of. I really appreciate you doing this work. Fixing the TLD issue is important. Thank you!

Perlmint:
Thanks for review. I'll push the fixed code soon.

About the testing, I have no actual idea. but it looks like consulting [the test data(provided by publicsuffix.org)](https://raw.githubusercontent.com/publicsuffix/list/master/tests/test_psl.txt) can give idea.

timofonic:
I have been banned from IANA and other shit because supossedly someone used
my IP address for spamming. Dynamic IP address are usually banned too.

I doubt this could be really a good effort to avoid spam and attacks, but
it's really an attack against home servers. They want all us to live in the
cloud.

willkg:
@timofonic I don't understand what you're getting at. Can you please explain more?

MadSpindel:
What is the status on this? Would be nice to have this merged into master.

willkg:
I had no idea @Perlmint updated this PR. When updating a PR branch, you need to comment otherwise maintainers aren't notified in any way.

I'll look at it when I get a chance. That'll probably be in February some time.

Perlmint:
I'm really sorry about I didn't any comment.


This is not fully working now. some special cases - should be not matched - are matched with generated RE.  
> `ac.jp`: should not matched as suffix. but It matched with 'jp' suffix.

but all cases should matched are correctly matched.


I tried to fix it, but I couldn't finished it because it is hard and I was suddenly busy with my job recent few weeks.


It looks like really hard or impossible to archive rules of public suffix with RE.  
I think matching url with simple way - e.g. re matching with generated from TLDs - and verifying matched url with non-re method is simple and works correctly. but it will be changing all parts using `url_re`.

willkg:
@Perlmint My apologies. I didn't mean to fault you, but what I wrote certainly indicates that. I think the fact that GitHub doesn't notify anyone when a PR branch is updated is incredibly frustrating. This issue comes up all the time. It's poor UX on their part.

About the rest of what you said, that's interesting. I don't know a lot about TLDs, so I'm not  sure I understand the issue. Is it the case that TLDs that are country codes require an additional part? e.g. `ac.jp` is bad, but `ac.co.jp` is ok?

Perlmint:
Oh,... I knew that github not notify when just commits are pushed. but I forgot about it. So I feel sorry about my forgetting about notification. :)


Your example is right. However TLDs are only one level. doesn't have additional parts. Problem I said is about the public suffix.  
At the beginning of this PR, I just added TLDs generated from [IANA TLD list](http://data.iana.org/TLD/tlds-alpha-by-domain.txt). and I recieved comment about using [Public suffix](https://publicsuffix.org/list/public_suffix_list.dat).  
Now, some domains are not valid in public suffix rules, but generated RE in my commit matches it.

willkg:
I'm sorry I've been letting this sit around so long. I've been wrestling with a few aspects of this PR and general direction:

1. If we land this or something like it, how do we maintain a list of TLDs going forward? Does that require us to do regular releases? Does that require us to do releases that are solely a TLD update? 
2. Are there security issues around TLDs that would be better served by separating maintenance of a list into a separate project?
3. This PR provides a command to update the TLD list, but we have no insight as to whether that operation is successful. It might be successful now, but what if it fails tomorrow? Will a user know that it's bad? Is there a way for a user to verify it? What's our responsibility for making sure it always works correctly?

I think requests has a similar situation where they used to have CA bundles in the requests package and then split that out into a separate project called certifi. Then certifi updates can come out independently of requests and it provides a nice API for people to use their own CA bundles.

I think treating this the same way is worth looking into. I want to take a look around to see what the current state of TLD-related Python libraries is and maybe we should use one of them.

I wrote up issue #223 to cover that work.

Perlmint:
I'm really sorry for my late response.

In my opinion, 
1. I have no idea about this. I think that the maintainers should choose suitable way for them.
2. A separate project will be better for maintenance.
3. Anyway, TLD list or public suffix list is too large to maintaining the rule by hand. It looks like need a generator. I have no nice idea about verification and responsibility.

Currently, this PR is not about TLD, but public suffix. The public suffix is too complicated(maybe or just I failed) to express with RE. I can't handle this PR - using RE. Actually, this is abandoned by me. So, I think this PR should be closed.
It will be better to forget this PR and treat TLD issue.

PR #162, opened 2015-09-09
Add skip_code option to linkify()

nikolas:
The markdown parser I'm using is rendering code blocks like
`<pre><code> ... </code></pre>` instead of `<pre> ... </pre>`.
This is breaking the `skip_pre` behavior, and attempting to linkify
URLs within my code blocks. Adding a skip_code option to linkify
and enabling it fixes my problem.

Let me know if you have any suggestions here.

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -116,7 +116,7 @@ def clean(text, tags=ALLOWED_TAGS, attributes=ALLOWED_ATTRIBUTES,
 
 
 def linkify(text, callbacks=DEFAULT_CALLBACKS, skip_pre=False,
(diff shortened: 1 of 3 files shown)


willkg:
I don't think I understand the issue here. Further, I'm pretty sure adding another skip argument is probably not what we'll want to do once we establish the details of the problem.

Can you open up an issue describing the problem and include a failing example or test? Then we can go from there.

Closing this out for now.

stfenjobs:
https://bleach.readthedocs.io/en/latest/linkify.html#skipping-links-in-specified-tag-blocks-skip-tags

PR #163, opened 2015-09-14
Add ve/ and venv/ dirs to gitignore

nikolas:
for virtualenv

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ dist
 .noseids
 build
 .tox
+ve/
+venv/


jsocol:
I strongly recommend using [virtualenvwrapper](https://pypi.python.org/pypi/virtualenvwrapper/4.7.1) which keeps the virtualenv directory outside the source dir, or using a profile-wide git-ignore file. I don't like including personal environmental config in project files.

nikolas:
Ah thanks, yeah I should have those directories in my global `.gitignore`.

PR #175, opened 2015-11-03
Remove dependency on six

kovidgoyal:
six is overkill for the basic compatibility requirements of bleach.

### DIFF
--- a/bleach/encoding.py
+++ b/bleach/encoding.py
@@ -4,9 +4,12 @@ import types
 import sys
 from collections import namedtuple
 if sys.version_info.major == 2:
-    six = namedtuple('Six', 'integer_types string_types text_type PY3')((int, long), (basestring,), unicode, False)
+    six = namedtuple('Six', 'integer_types string_types text_type PY3')(
+        (int, long), (basestring,), unicode, False)
 else:
-    six = namedtuple('Six', 'integer_types string_types text_type PY3')((int,), (str,), str, True)
+    six = namedtuple('Six', 'integer_types string_types text_type PY3')(
+        (int,), (str,), str, True)
+
 
 def is_protected_type(obj):
     """Determine if the object instance is of a protected type.
--- a/bleach/encoding.py
+++ b/bleach/encoding.py
@@ -1,8 +1,12 @@
 import datetime
 from decimal import Decimal
 import types
-import six
-
+import sys
+from collections import namedtuple
+if sys.version_info.major == 2:
+    six = namedtuple('Six', 'integer_types string_types text_type PY3')((int, long), (basestring,), unicode, False)
+else:
+    six = namedtuple('Six', 'integer_types string_types text_type PY3')((int,), (str,), str, True)
 
 def is_protected_type(obj):
     """Determine if the object instance is of a protected type.
--- a/bleach/encoding.py
+++ b/bleach/encoding.py
@@ -3,9 +3,9 @@ from decimal import Decimal
 import types
 import sys
 from collections import namedtuple
-if sys.version_info.major == 2:
+if sys.version_info[0] == 2:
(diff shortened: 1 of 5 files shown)


willkg:
I don't think using six is overkill since it's a very tiny library that's very good at what it does, is well documented, is well used by most libraries that maintain Python 2 and 3 compatibility (and thus is already installed) and is well maintained as new versions of Python come out. I'm loathe to copy the bits of six we use into bleach forcing us to maintain code that's already well maintained elsewhere.

Is the underlying issue a philosophical issue or do you have difficulties using six?

kovidgoyal:
The copied code is literally trivial, there is no "maintenance" involved, expecially since you run automated tests on all versions of python already.

Problems with six:

Personally:
1) I use bleach bundled in my project, which is distributed on many different platforms, the vast majority of which do not have six pre-installed, and I do not want to add a new dependency for such a simple use case. 

Philosophically:

1) In my experience the robustness of software is inversely correlated with the number of dependencies it has. The more moving parts, the more opportunities for breakage. And six is a completely unnecessary one, given that the things it is doing in bleach are so simple. 

2) There is a performance cost with using a *relatively* heavy weight solution like six just to test a few types. Indeed, while I did not do this in my PR, so as to keep code changes to a minimum, you should really re-write force_unicode to directly use global variables rather than a global variable + attribute lookup, since it is in the hot path. Depending on how hot, you could even make two versions of the function, py2 and py3 and install the correct version at runtime.

dstufft:
+1 on just continuing to use six. Dependencies, particularly pure Python ones, are not bad things to be avoided.

I'm also -1 on trying to get overly clever attempting to eek out small performance improvements in CPython. If someone needs more performance they should be using PyPy IMO.

kovidgoyal:
Really, avoiding unnecessary attribute accesses in the hot path is overly clever? And are you seriously suggesting that rather than profiling+optimizing code, telling people to use an alternate, incompatible interpreter is acceptable? 

Sure, why bother optimizing your own code, let someone else do it for you. Removing unnecessary attribute lookups is an optimization with no downside in terms of maintainability, code clarity, or any other metric you care to name -- but, so what, just use PyPy instead, CPU cycles are free.

jsocol:
Honestly, `bleach.encoding.force_unicode` is probably overkill, and it could probably largely be replaced with `six.u`.

> In my experience the robustness of software is inversely correlated with the number of dependencies it has. The more moving parts, the more opportunities for breakage.

My experience is the opposite (except NPM's community versioning issues). Every line of code is something that can be wrong or fail and must be tested. Let other people do it. Six has lots of eyes and a solid test suite.

As for optimizing attribute access: bleach is _slow_. Bleach uses html5lib which is a real parser: it reads and tokenizes input strings one character at a time. It does this in pure python. It builds and then _walks_ a tree in `linkify()`. It's almost certainly worth caching the output of bleach and not keeping it on a critical path, rather than worrying about micro-optimizations.

kovidgoyal:
And yet, there is a micro-optimization in `force_unicode` already. But, this is getting off the point of this PR. Please decide if getting rid of six is something you are willing to do or not. I will decide whether to use a private fork of bleach or not, accordingly.

willkg:
I think we should continue using six. html5lib also requires six, so I think we can ignore the other arguments for now:

https://github.com/html5lib/html5lib-python/blob/38baa0ef4cc8f15b0807c7c38720c60dd76badea/setup.py#L56

Given that, I don't think the changes herein make sense to do.

Thank you for looking into it!

willkg:
@kovidgoyal You included the bleach source code into calibre, but didn't add a note to the COPYRIGHT regarding bleach's license and copyright. Please add that.

kovidgoyal:
Sure, it will get done sometime before the next release.

PR #178, opened 2015-11-30
Hashtag support

defrex:
This adds hashtag support. It's not really useful without a callback to setup the path for the current application, since each will be different.

I needed it for my own purposes. Totally understand if there is no interesting in merging, just thought I'd throw it out there.

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -86,6 +86,8 @@ email_re = re.compile(
     """,
     re.IGNORECASE | re.MULTILINE | re.VERBOSE)
 
+hashtag_re = re.compile(r'#([\w\-_]+)', re.IGNORECASE | re.MULTILINE)
+
 NODE_TEXT = 4  # The numeric ID of a text node in simpletree.
 
 ETREE_TAG = lambda x: "".join(['{http://www.w3.org/1999/xhtml}', x])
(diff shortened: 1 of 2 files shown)


willkg:
I don't understand what this is. Can you explain the problem with relevant use cases?

defrex:
Sorry, perhaps I should have provided more details.

This 
```
here is my text #example #thisiswhatahashtagis
``` 
becomes: 
```
here is my text <a href="?tag=example">#example</a> <a href="?tag=example">#thisiswhatahashtagis</a>
```

In essentially the same way that email addresses are found and linkified this patch allows finding and linkifying hashtags.

Clearly `?tag=whatev` isn't going to help anyone, but it's a simple thing to add that application-specific code in a callback. I added an `_tag` attr to make that simpler.

willkg:
Ahh... got it! That makes sense and I understand the use case now.

I think there are a variety of ways to denote tags and users and other meaning in the data some of which can be linkified. I think these sorts of things are domain-specific and outside the scope of bleach. Similarly, I wouldn't want to implement auto-linking for reST links, Markdown links or github issues.

I think I'm -1 on this unless there's a very compelling reason why this needs to be part of bleach and shouldn't be elsewhere.

defrex:
Another approach to problems like this might be extensibility. The hashtag code was _very_ similar to the email code. It's basically just a regex to do the matching, and a function to convert the match object to an attrs dict.

If those could be passed into linkify generically: reST links, Markdown links, github issues, you name it, could all be accommodated. I wouldn't have needed to patch at all.

willkg:
I think given the last comment, this should get turned into an issue and discussed more there.

@defrex Can you write up an issue for this?

PR #188, opened 2016-02-02
Content of <pre> tags should not be linkified when skip_pre=True

### LINKED ISSUE #150: Children of <pre> tags should not be linkified when skip_pre=True (patch attached)
The children of `pre` tags should not be linkified when `skip_pre` is on 

```
diff --git a/bleach/__init__.py b/bleach/__init__.py
index 48b6512..4c2dd1b 100644
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -300,7 +300,7 @@ def linkify(text, callbacks=DEFAULT_CALLBACKS, skip_pre=False,
                 if node.tag == ETREE_TAG('pre') and skip_pre:
                     linkify_nodes(node, False)
                 elif not (node in _seen):
-                    linkify_nodes(node, True)
+                    linkify_nodes(node, parse_text)
 
             current_child += 1
 
diff --git a/bleach/tests/test_links.py b/bleach/tests/test_links.py
index 62da8d1..ae0fba7 100644
--- a/bleach/tests/test_links.py
+++ b/bleach/tests/test_links.py
@@ -314,6 +314,13 @@ def test_skip_pre():
     eq_(nofollowed, linkify(already_linked))
     eq_(nofollowed, linkify(already_linked, skip_pre=True))
 
+def test_skip_pre_child():
+    # Don't linkify the children of pre tags.
+    intext = '<pre><code>http://foo.com</code></pre>http://bar.com'
+    expect = '<pre><code>http://foo.com</code></pre><a href="http://bar.com" rel="nofollow">http://bar.com</a>'
+    output = linkify(intext, skip_pre=True)
+    eq_(expect, output)
+
 
 def test_libgl():
     """libgl.so.1 should not be linkified."""
```

ialbert:
Closes #150 and possibly #162

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -347,6 +347,13 @@ def linkify(text, callbacks=DEFAULT_CALLBACKS, skip_pre=False,
         if url.startswith('('):
             _wrapping = strip_wrapping_parentheses(url)
             url, open_brackets, close_brackets = _wrapping
+        elif url.endswith('/)'):
+            # This is a very common corner case where the URL is listed
+            # last within a longer text within parentheses.
+            # Example: (check my site at http://foo.bar/)
+            # Do not consider the closing ) as part of the URL.
+            url, open_brackets, close_brackets = url[:-1], 0, 1
+
         end = ''
         m = re.search(punct_re, url)
         if m:
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -318,7 +318,7 @@ def linkify(text, callbacks=DEFAULT_CALLBACKS, skip_pre=False,
                 if node.tag == ETREE_TAG('pre') and skip_pre:
                     linkify_nodes(node, False)
                 elif not (node in _seen):
-                    linkify_nodes(node, True)
+                    linkify_nodes(node, parse_text)
 
             current_child += 1
 
--- a/bleach/tests/test_links.py
+++ b/bleach/tests/test_links.py
@@ -470,4 +470,3 @@ def test_skip_pre_child():
     expect = '<pre><code>http://foo.com</code></pre>'
     output = linkify(intext, skip_pre=True)
     eq_(expect, output)
-
--- a/bleach/tests/test_links.py
+++ b/bleach/tests/test_links.py
@@ -462,3 +462,11 @@ def test_remove_first_childlink():
     callbacks = [lambda *a: None]
     eq_(expect,
         linkify('<p><a href="/foo">something</a></p>', callbacks=callbacks))
+
+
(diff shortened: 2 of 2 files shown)


ialbert:
New addition also closes #190

bleach/__init__.py
```diff
@@ -347,6 +347,13 @@ def link_repl(match):
         if url.startswith('('):
             _wrapping = strip_wrapping_parentheses(url)
             url, open_brackets, close_brackets = _wrapping
+        elif url.endswith('/)'):
+            # This is a very common corner case where the URL is listed
+            # last within a longer text within parentheses.
+            # Example: (check my site at http://foo.bar/)
+            # Do not consider the closing ) as part of the URL.
+            url, open_brackets, close_brackets = url[:-1], 0, 1
```
willkg: This fix isn't related to the rest of this PR. I'm not sure I understand the issue here and I'm not wild about this fix either. If I do understand the issue correctly, seems like this is a problem with the regex and should be dealt with there.

I think it's probably worth opening an issue for this if there isn't one already so we can figure that out separately and removing this bit from the PR.

willkg:
It's a little frustrating that the bulk of the changes in this PR are style changes that have nothing to do with the purpose of the PR. That makes it harder to review. In the future, please don't do that.

This PR seems to solve two different issues. I think we should split out the second one--I said as much in the comment.

Other than that, the changes in this PR seem ok. I have to think through the specific problem and how this change affects other situations.

bleach/__init__.py
willkg: I see what happened. You wrote up the issue and then tossed something more in this PR. It's a lot easier to work through pull requests when they don't have a pile of changes in them.

ialbert: Sorry about that, PyCharm did that automatically and I did not realize that it reformatted lines. 

We had this discussion before and it is just another example how using flake8 and its ilk to enforce the code "quality" makes life more miserable for everyone. Basically it is counterproductive.

Tests kept failing because the buildbot would enforce rules for code not being formatted right, but then when one uses a formatter that is supposedly pep8 aware and makes these errors go away it ends up reformatting the code with rules that flake8 did not enforce. 

By the time of the third attempt to fix failing test that had nothing to do with code quality I lost the focus of ensuring that the changes are minimal and just wanted to shut it up already.

willkg:
I created PR #202 to cover the ) issue.

I created PR #205 to cover the skip_pre issue.

That covers everything in this PR, so I'm going to close this out.

ialbert:
Thanks @willkg much appreciated!

PR #195, opened 2016-05-18
Limit html5lib version before sanitizer API changes get released

gsnedders:
Now https://github.com/html5lib/html5lib-python/pull/110 has landed, Bleach needs to limit its version of html5lib until it's changed to follow the new API. (This is likely going to be the last breaking API change that will affect Bleach until 2.0.)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 ordereddict
 six
-html5lib>=0.999
+html5lib>=0.999,<0.99999999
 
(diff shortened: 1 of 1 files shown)


willkg:
Thanks for the heads up!

gsnedders:
FWIW, I expect a release within a fortnight.

willkg:
Thank you!

Issue #197, opened 2016-05-23
unify version information


willkg:
Currently, the version is in three separate files:

1. `setup.py`
2. `docs/conf.py`
3. `bleach/__init__.py`

That's annoying plus if you look at past releases, it's clear we miss one periodically.

This bug covers moving the version into `bleach/__init__.py` and changing the other two to look there for the version.

TimDumol:
I tried tackling this here: https://github.com/mozilla/bleach/pull/198. Hope it's good enough!

willkg:
This was fixed with #198. Closing it out. Thank you @TimDumol !

PR #198, opened 2016-05-27
Unify version information to rely on a single source.

TimDumol:
For issue #197. I used a separate `version.py` (as per http://stackoverflow.com/a/24517154/112943) since directly importing `__init__.py` doesn't work if you don't have dependencies installed.


setup.py
```diff
@@ -20,10 +21,17 @@ def get_long_desc():
     desc += open('CHANGES').read()
     return desc
 
+# foolproof way of single-sourcing version as per
+# http://stackoverflow.com/a/24517154/112943
+version_namespace = {}
+version_path = convert_path('bleach/version.py')
+with open(version_path) as version_file:
+    exec(version_file.read(), version_namespace)
```
willkg: I'm not wild about doing an `exec` in `setup.py`. Generally how I deal with this is by parsing out the line with `__version__` in it.

https://github.com/willkg/dennis/blob/master/setup.py#L19

Can you redo this using the same method? The version can stay in `bleach/version.py` if you want--I don't really care so much about that aspect.

willkg:
Thank you for working on this! I really appreciate it!

setup.py
TimDumol: I actually had similar thoughts about this. `exec` doesn't feel right, especially in a security package. Will do.

TimDumol: Fixed! Also, there seem to be some build failures due to `flake8` that affect recent PRs. Shall I include the `# flake8: noqa` comment to make it pass here too?

setup.py
```diff
@@ -20,10 +23,24 @@ def get_long_desc():
     desc += open('CHANGES').read()
     return desc
 
+# foolproof way of single-sourcing version as per
+# http://stackoverflow.com/a/24517154/112943
```
nickburlett: This comment is now out of date. The method at the link uses `exec` but this code no longer does.

TimDumol: While it doesn't use the exact code, I feel like it still uses the same general idea (separate file and reading it without importing it, as opposed to using `setuptools` or trying to import the file somehow). If you still feel this comment does not apply, I don't mind removing it.

nickburlett: @TimDumol : Thinking about it more, I think the link is worth keeping, but I'd add something about why it isn't done in exactly he same way. Perhaps:

```python
# Foolproof way of single-sourcing version inspired
# by http://stackoverflow.com/a/24517154/112943
# Here we use re.search instead of exec to avoid any
# possibility of side effects in version.py
```

It's probably not super-important, I've just been thinking about code commenting a lot lately. I prefer documenting the reason for doing something in addition to where the idea came from.

TimDumol: Well, I agree with your suggestion, so I'll go through with it :D I agree that comments are important, and especially the absence of misleading comments.

willkg:
Awesome! Thank you!

PR #199, opened 2016-05-27
Add param to remove contents inside a disallowed tag

### LINKED ISSUE #185: Add option to remove content inside of stripped tags
Currently, if you have a block of html with a javascript `<script>` tag within it, you can strip the _tag itself_, but not the actual javascript.

```
>>> import bleach
>>> test = '<div><p>A test</p><script>console.log("test")</script></div>'
>>> bleach.clean(test, strip=True)
u'A testconsole.log("test")'
```

It would be nice if we could strip all of the javascript out of the text entirely.

aysonje:
closes #185

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -17,7 +17,8 @@ class BleachSanitizerMixin(HTMLSanitizerMixin):
     allowed_svg_properties = []
 
     def _has_unmatched_tag(self):
-        return self.strip_disallowed_element_contents and len(self.unmatched_tags) > 0
+        return (self.strip_disallowed_element_contents
+                and len(self.unmatched_tags) > 0)
 
     def sanitize_token(self, token):
         """Sanitize a token either by HTML-encoding or dropping.
@@ -37,7 +38,8 @@ class BleachSanitizerMixin(HTMLSanitizerMixin):
 
         if token['type'] in (tokenTypes['StartTag'], tokenTypes['EndTag'],
                              tokenTypes['EmptyTag']):
-            if token['name'] in self.allowed_elements and not self._has_unmatched_tag():
+            if (token['name'] in self.allowed_elements and
+                    not self._has_unmatched_tag()):
                 if 'data' in token:
                     if isinstance(self.allowed_attributes, dict):
                         allowed_attributes = self.allowed_attributes.get(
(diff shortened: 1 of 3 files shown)


calvin:
The following fails:
```python
html = '''
<script type="text/javascript">// <![CDATA[
console.log('<a class="" href="" target=""><i class=""></i></a><a class=""><i class=""></i></a>');
// ]]></script>
'''
bleach.clean(html, strip=True, strip_contents=True)

html = '''
<script type="text/javascript">// <![CDATA[
console.log('<a class="" href="" target=""><i class=""></i></a><a class=""><i class=""></i></a>');
// ]]></script>
'''
bleach.clean(html, strip=True, strip_contents=True)
```

Expected outputs: string with whitespace characters

bleach/sanitizer.py
```diff
@@ -16,6 +16,9 @@ class BleachSanitizerMixin(HTMLSanitizerMixin):
 
     allowed_svg_properties = []
 
+    def _has_unmatched_tag(self):
+        return self.strip_disallowed_element_contents and len(self.unmatched_tags) > 0
+
     def sanitize_token(self, token):
```
calvin: I think _L40_ should be `if token['name'] in self.allowed_elements and not self._has_unmatched_tag():`. What do you think?

aysonje: You're right. I missed that part. Thanks!

calvin:
Could you address CI lint errors? You can find them on https://travis-ci.org/mozilla/bleach/jobs/133921872#L277-L286 since I'd like to see this changeset merged for the next release.

willkg:
I haven't had a chance to look through this, yet. I put it in the v1.5 milestone and will work through it next week.

willkg:
I've thought about this for a while and I think I'm going to pass on it for 1.5.

There are two big reasons for doing this:

1. `.clean()` is about removing malicious content--not about transforming HTML documents for other mediums or prettifying content. `.clean()` is a security-focused function and as such, keeping its functionality minimal reduces the likelihood of bugs that have security-related impact. That's really important.
2. Seems like the use cases for stripping content in tags are related to transforming HTML documents and prettifying content. I think that should get handled by a different function or possibly not by bleach at all.

Towards that, I've clarified the goals/non-goals language and updated the docs to make it clearer what `.clean()` is supposed to be doing and what it's not going to be good for.

Given that, I'm going to pass on this and close it out.

I do appreciate the work that was done here. I'm game for talking about building a `.prettify()` function which is about prettifying HTML which would cover the use these changes cover.

PR #202, opened 2016-06-14
Fix linkify edge case with ending )

### LINKED ISSUE #190: Linkify should not match the closing parenthesis right after the TLD
Linkify handles adjacent parentheses correctly for example this will work 

    (http://bar.com)

the link will go to http://bar.com

But adding a word right after the opening parenthesis will break the linking. For example when linkifying

    (see  http://bar.com/)

the  href will contain the closing parenthesis ` )` as well. Technically this is correct as the closing parenthesis is a valid construct there, in practice this is never used this way.

When accepting free user input it is surprisingly common to have users  type links in the format that I mention, and it always results in invalid links. Note how just about all autolinkers (github's included correctly recognize this situation)

IMO linkify should either refuse to linkify this text or match it correctly because what it does is never correct regarding what the user intent actually was. Moreover it is very difficult to catch this error visually.

willkg:
It's common to put urls in parentheses. Sometimes the url_re will pick
up end parentheses that are part of the parenthetical, but shouldn't be
part of the url. This fix alleviates that somewhat by checking to see if
the url has a ( and if not, stripping any ) from the end. This assumes
that urls won't end in ) without also having a ( in them.

This is based on a fix from Istvan Albert.

Fixes #190 

r?


willkg:
I'm going to self-r+ this. Seems good. We can back it out and/or enhance the fix later if it proves problematic.

PR #183, opened 2015-12-04
Added NullHandler to logging

dangayle:
Added NullHandler to logging, to prevent "No handlers could be found for logger "bleach"" warnings in applications that haven't set up logging properly. 

Fixes #182

PR #203, opened 2016-06-14
182 nullhandler

willkg:
This picks up the commit from #183 and adds another commit that fixes the issue that was having with Travis.

r?


willkg:
This seems good, so I'm going to self-r+ this.

PR #207, opened 2016-06-18
Use ASCII digits in port number parsing

Zopieux:
Currently, bleach accepts (linkyfies) the following “URLs”:

- `http://foo.com:𝟠𝟘𝟠𝟘/`
- `http://foo.com:٣٩٩٩/`

That is because the `\d` modifier of Python unicode regular expressions matches all digits in the Nd block [search for `category [Nd]` here](https://docs.python.org/3.5/library/re.html#regular-expression-syntax), and there are [many things](http://www.fileformat.info/info/unicode/category/Nd/list.htm) in there.

This PR replaces `\d` with `[0-9]+` and two test URLs to assert these links are not recognized.

--
Note: an alternative would be to use the `re.ASCII` flag but:
- it's hard to be compatible with Python 2
- it may not be appropriate because we *do* want to match funky unicode chars in other parts of the URL

--
Note: funnily enough, GitHub-flavored markdown also linkyfies them! http://foo.com:𝟠𝟘𝟠𝟘/

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -71,7 +71,7 @@ PROTOCOLS = HTMLSanitizer.acceptable_protocols
 url_re = re.compile(
     r"""\(*  # Match any opening parentheses.
     \b(?<![@.])(?:(?:{0}):/{{0,3}}(?:(?:\w+:)?\w+@)?)?  # http://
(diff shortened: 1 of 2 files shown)


willkg:
Tossing this in the v1.5 milestone. I'll look through it next week.

Issue #212, opened 2016-07-15
bleach breaks because of html5lib


matthiaskubik:
Since a few hours, bleach is broken because of a new version of html5lib:
 
File "/<sanitized>/endpoint/rest_endpoint_handler.py", line 21, in <module>
    import bleach
File "/usr/local/lib/python2.7/site-packages/bleach-1.4.2-py2.7.egg/bleach/__init__.py", line 8, in <module>
    from html5lib.sanitizer import HTMLSanitizer
ImportError: No module named sanitizer

Fix would be documented here https://github.com/html5lib/html5lib-python/issues/277

sergio97:
Looks like html5lib move/removed 'sanitizer'. This was intentional it seems

generalov:
> 0.99999999/1.0b9
> Released on July 14, 2016
> 
> Get rid of the sanitizer package. Merge sanitizer.sanitize into the sanitizer.htmlsanitizer module and move that to saniziter. This means anyone who used sanitizer.sanitize or sanitizer.HTMLSanitizer needs no code changes.

Unfortunally, bleach is using `from html5lib.sanitizer import HTMLSanitizer`.

jonathanmorgan:
Any idea of time frame on figuring out a fix?  If it is a big change and will take time, I understand, just want to get an idea.

matthiaskubik:
I already opened a pull request to fix the requirements.txt file, but no respose or action yet.
At the end, the requirements.txt has the wrong version: It has the one with 8 nines, where it should have the one with one 7. All we need is someone accepting the pull request as a quick workaround.

willkg:
I can't reproduce this issue. Here's my attempt:

```
$ mkvirtualenv bleachtest
<testing> $ pip install bleach
Collecting bleach
  Downloading bleach-1.4.3-py2-none-any.whl
Collecting six (from bleach)
  Using cached six-1.10.0-py2.py3-none-any.whl
Collecting html5lib<0.99999999,>=0.999 (from bleach)
Installing collected packages: six, html5lib, bleach
Successfully installed bleach-1.4.3 html5lib-0.9999999 six-1.10.0
<testing> $ python
Python 2.7.12 (default, Jul  1 2016, 15:12:24) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import bleach
>>> from html5lib.sanitizer import HTMLSanitizer
>>> 
```

That seems to work fine to me.

[...]

PR #213, opened 2016-07-18
changed max prereq to 0.9999999 as 0.99999999 was created a couple of…

matthiaskubik:
… days ago and was even not backward compatible so that bleach breaks. See https://github.com/mozilla/bleach/issues/212


matthiaskubik:
obsolete

PR #214, opened 2016-07-21
Pin down the version of html5lib

AntonNguyen:
Version `0.999999999` has some breaking changes:

```
   File "/var/lib/jenkins/workspace/app/env/lib/python2.7/site-packages/bleach/__init__.py", line 8, in <module>
     from html5lib.sanitizer import HTMLSanitizer
 ImportError: No module named sanitizer
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 ordereddict
 six
-html5lib>=0.999,<0.99999999
+html5lib==0.9999999
 
 # Requirements to run the test suite:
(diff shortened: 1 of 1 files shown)


willkg:
The requirements file already correctly restricts the maximum version of html5lib to versions that bleach works with. Your change restricts it to a specific version which isn't what we want to do since we want to support multiple versions of html5lib.

Given that, I'm going to close this out without merging it.

PR #215, opened 2016-07-22
html5lib moved location of sanitizer for ver 0.99999999 and up

haganf:
I was getting an error with bleach with the latest version of html5lib, version 0.99999999. Appears the import for HTMLSanitizer has moved locations. Not sure if this is the best way to fix this but worked with 0.9999 and 0.999999999.

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -11,7 +11,10 @@ except ImportError:
 import re
 
 import html5lib
-from html5lib.sanitizer import HTMLSanitizer
+try:
+    from html5lib.filters.sanitizer import HTMLSanitizer
+except ImportError:
+    from html5lib.sanitizer import HTMLSanitizer
 from html5lib.serializer.htmlserializer import HTMLSerializer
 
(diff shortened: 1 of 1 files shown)


bleach/__init__.py
```diff
@@ -9,9 +9,13 @@ class NullHandler(logging.Handler):
         def emit(self, record):
             pass
 import re
+from distutils.version import LooseVersion
 
 import html5lib
-from html5lib.sanitizer import HTMLSanitizer
+if(LooseVersion(str(html5lib.__version__)) < LooseVersion('0.99999999')):
```
dstufft: I think it'd be better to just do:

```python
try:
    from html5lib.filters.sanitizer import HTMLSanitizer
except ImportError:
    from html5lib.sanitizer import HTMLSanitizer
```

haganf:
That did look a bit better, here's a different patch using the try/except instead.

haganf:
Aborting this PR, looks like html5lib is changing a lot... so this PR isn't necessary..

Issue #204, opened 2016-06-14
switch to py.test


willkg:
There are a handful of tests that are "interesting" and would benefit greatly from switching to py.test and taking advantage of py.test things.

Given that I'm maintaining this now, I want to switch to py.test and make my life easier.

This issue covers that.

willkg:
PR #222 landed, so I'm closing this out.

Issue #206, opened 2016-06-16
drop support for python 2.6


willkg:
Python 2.6 is done. Time to drop support for it and move on with our lives.

This issue covers removing the bits that test in Python 2.6 and any code that's specifically there to support Python 2.6.

willkg:
Note to self: we can remove ordereddict from the requirements file.

PR #222, opened 2016-10-31
Pytest rewrite

willkg:


willkg:
This fixes issue #204.

willkg:
This failed in Python 2.6 because flake8 doesn't support 2.6 anymore. Issue #206 covers dropping support for Python 2.6, so I'm not going to worry about this.

This failed in Python 3.2 because py.test doesn't support Python 3.2. I don't think we should be supporting 3.2 anymore since it was end-of-lifed earlier this year. I'll write up an issue about that.

This passed everywhere else.

willkg:
This seems fine. If there are issues, we'll bump into them. Given that, I'm going to self-r+ it and merge it.

PR #225, opened 2016-10-31
Use ASCII digits in port number parsing

willkg:
This fixes PR #207 which broke when I converted to py.test.


willkg:
Same issue as the previous PR I just landed: failed with Python 2.6 because flake8 doesn't support it and failed with Python 3.2 because py.test doesn't support it. I'll address that next, but I'm going to merge this now anyhow.

willkg:
@Zopieux Thank you!

Zopieux:
My pleasure.

PR #226, opened 2016-10-31
Allow travis to fail with Python 2.6 and 3.2

willkg:
It's failing now with Python 2.6 because flake8 doesn't work with Python
2.6. It's failing with Python 3.2 because py.test doesn't work with
Python 3.2.

I'm not quite ready to drop support for both of them, though. So for
now, let's allow the failures.


willkg:
Yay for failures!

PR #227, opened 2016-10-31
Fix test_idempotent to not sometimes fail

### LINKED ISSUE #161: test_basics:test_idempotent non-deterministically fails
Originally reported here: https://bugs.debian.org/798441

```
$ while nosetests3 bleach.tests.test_basics:test_idempotent; do :; done
```

This eventually results in https://gist.github.com/lamby/bacb42d3fffde936d4a6

Curiously it only occurs under Python 3, ie. not if we `s/nosetests3/nosetests/`

willkg:
linkify can return text with html attributes in a different order
depending on how the attributes come out of the attrs dict. Because of
that, there are several possible outcomes.

This fixes the linkify test to accept both of them.

Fixes #161


willkg:
Pretty sure that will fix it. Merging.

PR #219, opened 2016-09-09
Do not add trailing period when email address is at the end of a sentence

znerol:

PR #228, opened 2016-10-31
Do not add trailing period when email address is at the end of a sent…

willkg:
This fixes PR #219 which I broke when I pushed py.test rewrite.


willkg:
Seems fine, though I'm not positive I understand the consequences of removing that `\.` at the end of that complex and not particularly well documented url regexp.

PR #233, opened 2016-11-02
Clarify docs regarding appropriate use cases and goals

willkg:
This clarifies some of the guiding principles behind what bleach is for
and not for so it's clearer to users whether their needs will be met.

### DIFF
--- a/docs/goals.rst
+++ b/docs/goals.rst
@@ -45,7 +45,7 @@ content, and will use the HTML5 parsing algorithm to handle invalid markup.
 See the :ref:`chapter on clean() <clean-chapter>` for more info.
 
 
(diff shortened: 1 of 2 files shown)


docs/goals.rst
```diff
@@ -62,18 +64,43 @@ Non-Goals
...
+
+
+Clean up after trusted users
+----------------------------
+
+Bleach is powerful but it is not fast. If you trust your users, trust them and
+don't rely on Bleach to clean up their mess.
+
+
+Make malicious content look pretty or sane
```
jsocol: 👏

willkg:
Thank you @jsocol !

PR #240, opened 2016-12-08
Remove support for Python 2.6 and 3.2

### LINKED ISSUE #206: drop support for python 2.6

### LINKED ISSUE #224: drop support for python 3.2
https://www.python.org/dev/peps/pep-0392/#lifespan

Python 3.2 was end-of-lifed in February 2016. The version of py.test that we're switching to doesn't support it. It's worth dropping support for it.

willkg:
Fixes #206
Fixes #224

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,7 @@ cache:
   directories:
   - "~/.cache/pip"
 python:
-- "2.6"
 - "2.7"
-- "3.2"
 - "3.3"
 - "3.4"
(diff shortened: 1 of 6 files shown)


willkg:
I think this looks super duper. Landing it.

Issue #66, opened 2012-06-27
add tests for XSS examples in OWASP testing guide


graingert:
Adding all the tests from http://ha.ckers.org/xss.html would be a good idea

willkg:
The site in question isn't available anymore. Looks like some/all of the content was moved to the OWASP testing guide, so I grabbed the stuff in the appendix there.

Having said that, I'm not sure what "enough" entails, so I'm going to change the scope of this to "add tests for the XSS examples listed in the OWASP guide".

graingert:
@willkg sucky, I remember back in 2012, http://ha.ckers.org/xss.html was pretty great.

graingert:
@willkg here's the last good snapshot: https://web.archive.org/web/20120717070023/http://ha.ckers.org/xss.html

graingert:
@willkg and I've got a snapshot on github https://github.com/graingert/xss

willkg:
@graingert Can you go through and add those? If you use the system I used for regression tests, it's not too hard.

graingert:
Sorry I'm way more busy than I was 4 years ago!

willkg:
I added some stuff a while back. We can add new tests as needed. I don't think we need to keep this open forever.

PR #241, opened 2016-12-19
Add regression tests

willkg:
We're getting ready to overhaul bleach to use html5lib 0.99999999+.
Before doing that, I wanted to add more regression tests so I can verify
when/how things change. This adds those.

These tests come from the OWASP Testing Guide 4.0 appendix in the Cross
Site Scripting section.


willkg:
I have no idea if this covers issue #66 or not because I don't know what "enough" entails, but it does add a series of XSS tests.

These are added as regression tests to make sure `bleach.clean` doesn't vary between versions of bleach, Python and html5lib.

willkg:
It'd be great if other people could go through the tests and verify the output is correctly sanitized. I'll write that up as an issue.

Merging this now.

Issue #76, opened 2012-09-23
Add the ability to restrict/rewrite URLs for href and src attributes


lieryan:
Sample use cases: in moderated sites, linking to externally-hosted images means the user can change image without the moderator's knowledge. The ability to rewrite URLs means these images can either be sanitized or copied locally (like StackOverflow did with imgur). It can also rewrite absolute urls to relative urls.

[List of HTML tag attributes which can contain URLs](http://stackoverflow.com/questions/2725156/complete-list-of-html-tag-attributes-which-have-a-url-value)

EmilStenstrom:
Hi! This should definately be a priority in my opinion. Allowing <img> tags to have src attributes, or <a> tags to have href attributes, opens you up to all sorts of XSS problems, since you don't by default validate that the link or src points to a valid URL. 

Validating that the link is a valid URL (with one of the whitelisted protocols) would be a great first step imho.

EmilStenstrom:
Here's an example:

```html
python -c "import bleach; print bleach.clean('<img src=\"</comment><img src=x onerror=alert(1)//\">', strip=True, tags=['img'], attributes={'img': ['src']})"
```
The output is:
```html
<img src="</comment><img src=x onerror=alert(1)//">
```
... which is clearly not sanitized.

jsocol:
@EmilStenstrom If you think something is a functional attack against Bleach, please don't post examples in public. I'll get a better security-issue reporting email address into the readme/contributing docs today.

This, however, is not. If you plug this into a browser, there's no alert:

```html
<!DOCTYPE html>

<img src="</comment><img src=x onerror=alert(1)//">
```

That's because [attribute values can contain all those characters](http://www.w3.org/TR/html5/syntax.html#attributes-0). Attribute values aren't parsed. That HTML asks the browser to load the path `</comment><img src=x onerror=alert(1)//`, which is very nearly a valid path, too:

<img width="520" alt="image of the request in the firefox dev tools network panel" src="https://cloud.githubusercontent.com/assets/180459/9330270/41c72b64-4586-11e5-8dbc-97118c960e45.png">

Second, handling this example is already possible via [callables](http://bleach.readthedocs.org/en/latest/clean.html#callable-filters), which were documented in #77 (d61e8ee3ba3382d08ed236e7e402d048ac8179c4). They were added with the _explicit_ purpose of imposing limits on `src` attributes, like "must be from this (set of) domain(s)" or "must be relative". Or "can be parsed as a URL or path at all".

[...]

Issue #58, opened 2012-03-06
A cleaning pipeline


originell:
One thing that occured to me while working on the `BleachSanitizerMixin.sanitize_token` is that it's code is getting pretty long and kind of ugly, the more functionality one needs do add.

One possible way to counter this problem might be to implement a "pipeline". I've seen this concept first in [django-social-auth](https://github.com/omab/django-social-auth) which I'm also using in a project. To be more specific [this part of the README](https://github.com/omab/django-social-auth#id14) explains it pretty good. Personally I really love the concept. The code of the [default social auth pipeline is here](https://github.com/omab/django-social-auth/tree/master/social_auth/backends/pipeline)

So what I am proposing is something very much in the vain of social auth's pipeline:
1. Instead of having to tie into the (big) if condition, we iterate over an iterable with functions::
   
   ``` python
   PIPELINE = ('SkipAllowedElements', 'StripScripts',...)
   ```
2. While iterating, each function get's called and returns either a cleaned token which gets passed on to the next function in the iterable _or_ `None` for skip. Note that maybe it might be more expressive to use Exceptions (like `SkipToken`) here.

This may also clean up a bit of the code I introduce with the `strip_script_content` fork (like doing `self.previous_token = token` right before every `return`)

So, basically this is me requesting for a comment on refactoring things a bit :D – Maybe the same kind of _pipeline-logic_ could be used for #56

willkg:
In html5lib >= 0.99999999, sanitizing happens as a filter after tokenizing and you can easily add additional filters. With the rewrite for Bleach 2.0, you can trivially use the `BleachSanitizerFilter` with other html5lib filters (and ones you write yourself) for a cleaning pipeline.

After I finish the rewrite, I'll verify that's true and add an item to the docs about it.

PR #243, opened 2016-12-31
Support linkify() callbacks to replace attributes for tags other than <a>

rshk:
This will allow, for example, to replace URLs in 

See mozilla/bleach#76

In the long term, it might be worth refactoring linkify() a bit, to reduce complexity in that single function and make it more extensible instead.. but I think this should do for now. Comments?

Also, happy new year! 🎉


willkg:
This is interesting, but not in the scope of `linkify`. I'm not entirely sure it's in the scope of `clean` either, but now that Bleach uses html5lib 0.99999999 and sanitizing is just a filter, we can take advantage of the filtering API and these sorts of things can be solved in that manner.

That's covered in #58.

Given that, I'm going to close this out. I'll add a note in #76 about it.

PR #250, opened 2017-02-22
Rewrite to use html5lib >= 0.99999999

### LINKED ISSUE #229: rewrite bleach to support html5lib >= 0.99999999 (8 9s)
I thought someone was going to write up an issue for this, but apparently no one has, yet.

html5lib changed some APIs so we pinned to specific versions of html5lib with some number of 9s.

html5lib 1.0 milestone is pretty close to done. I think they're just waiting on documenting their public APIs. I don't see any indication of what their schedule looks like, but from what I did see, it seems like it's "done" except for the documentation. Given that, we could probably be safe by adding support for html5lib 0.999999999 (that's 9 9s) in addition to the ones we support now.

There's another issue in our tracker about how 0.999 causes some errors. We might want to nix support for that version.

This issue covers adding support for html5lib 0.999999999 (that's 9 9s) and 1.0 when that releases.

willkg:
This needs another going over. Plus I bumped into a few areas that don't have behavior tests, so it's possible there are things that are broken or not finished, yet. I'll keep working on those.

Fixes #229


willkg:
What's next:

1. implement missing tests
2. review the changes one more time and make sure `CHANGES` is correct
3. review docs
4. put this in a `.tar.gz` and host somewhere for others to test
5. fix any issues that come out of that
6. land this

tests/test_security.py
```diff
@@ -88,8 +90,11 @@ def test_invalid_filter_attr():
 
 def test_invalid_tag_char():
     assert (
-        clean('<script/xss src="http://xx.com/xss.js"></script>') ==
-        '&lt;script xss="" src="http://xx.com/xss.js"&gt;&lt;/script&gt;'
+        clean('<script/xss src="http://xx.com/xss.js"></script>') in
+        [
+            '&lt;script src="http://xx.com/xss.js" xss=""&gt;&lt;/script&gt;',
+            '&lt;script xss="" src="http://xx.com/xss.js"&gt;&lt;/script&gt;'
```
willkg: I think to fix this we need to sort the attributes and then escape it. That happens in `.disallowed_token` in the sanitizer. I'll think about that.

willkg:
Ok. I think that's good enough for now. Merging.

Issue #242, opened 2016-12-19
verify sanitizated output for regression tests


willkg:
In pull #241, I added regression tests for `bleach.clean`. We should verify the output for those regression tests.

willkg:
Tagging @g-k with this one.

g-k:
👍 I'm planning to pull in the test page from https://cure53.de/purify and modify it to POST to a test server that calls `bleach.clean` and plugs the result back into the DOM.

Hopefully we can reuse their karma test runner to easily check against browsers too: https://github.com/cure53/DOMPurify/blob/master/test/karma.conf.js

willkg:
The output is probably fine.

PR #251, opened 2017-02-23
add test website and scripts

g-k:
refs: https://github.com/mozilla/bleach/issues/242 https://github.com/mozilla-services/foxsec/issues/133

This probably shouldn't go in as it stands, but I thought I'd document my changes here.

The test page was initially copied from https://cure53.de/purify (dual MPL 2 and Apache licensed which should be compatible) and requires some attribution.

glorious web 1.0 styles:

screen shot 2017-02-22 at 7 41 58 pm

### DIFF
--- a/website/.gitignore
+++ b/website/.gitignore
@@ -1,0 +1,1 @@
+testcases.json
--- a/website/README.txt
+++ b/website/README.txt
@@ -1,0 +1,16 @@
+Scripts for a Bleach demo/test website
+
+Usage:
+
+from the project root:
+
+# generate testcases.json
+python website/data_to_json.py tests/data > testcases.json
+
+# run the test server
+cd website && python server.py &
+
+# open the page in browsers python can find
+python open_test_page.py
+
+# inspect bleached html and iframe
--- a/website/data_to_json.py
+++ b/website/data_to_json.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+
+"""
+Util to write a directory of test cases with input filenames
+<testcase>.test and output filenames <testcase>.test.out as JSON to
+stdout.
+
+example:
+
+python tests/data_to_json.py tests/data > testcases.json
+"""
+
+import argparse
+import fnmatch
+import json
+import os
+import os.path
+
+import bleach
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument('data_dir',
+                        help='directory containing test cases with input files'
+                        ' named <testcase>.test and output <testcase>.test.out')
+
+    args = parser.parse_args()
+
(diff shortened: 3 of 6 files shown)


willkg:
This looks really interesting and I <3 how this builds a framework we can reuse going forward. I have a few thoughts on this.

It'd be good to be able to verify that the test harness is working by using one of those test cases without passing it through `bleach.clean` and having it do a JS alert or something like that. Maybe have an additional button that does what the `onload` event listener is doing, except not running it through `bleach.clean`?

Can we say with 100% confidence that running all the tests through `bleach.clean` and concatenating the output and displaying that is equivalent to displaying them one at a time? In other words, by concatenating them, do we inadvertently reduce the efficacy of the test?

g-k:
> Maybe have an additional button that does what the onload event listener is doing, except not running it through bleach.clean?

+1  I tried this by commenting out the code a button to do it would be better.  

> Can we say with 100% confidence that running all the tests through bleach.clean and concatenating the output and displaying that is equivalent to displaying them one at a time? 

No, that's just an artifact of the page I pulled in. I could see dangling markup vectors might somehow canceling each other out.

> In other words, by concatenating them, do we inadvertently reduce the efficacy of the test?

Yes.

Improvements TODO:

- [ ] button to pass a testcase through without cleaning
- [ ] testcase isolation: create one iframe per testcase or iterate through sending them to one iframe
- [ ] add an endpoint for testing linkify?

Also, instead of manually validating each testcase, I was thinking we could instrument the page to check things like window.alert is called on the dirty output and not called on the bleached output.

willkg:
I haven't wrapped my head around linkify, yet, so I don't know what the deal is there. With `clean`, we're making security-related promises about the output. I'm not sure we're doing the same with `linkify`. But maybe we are?

g-k:
The linkify docs say there's some sanitization, but I'm not sure how that's implemented.

> By default, linkify() will perform some sanitization, only allowing a set of “safe” tags.

Going to punt on that for now.

willkg:
I'll look at linkify next week. Maybe it's due for an overhaul especially if we're making safety promises. Part of me wonders if it's doing sanitization, whether we should just make it a flag for clean. It's probably just a Filter at this point.

willkg:
@g-k Is this waiting on anything or is it ready to get reviewed again?

g-k:
It's ready to be reviewed again.

willkg:
Ok! Sorry about that. I didn't mean to let this sit for so long.

I'll look through it some time this week again and land it.

g-k:
No problem! I'm open to other approaches like putting it in a different
library or closing this in favor of something else too.

willkg:
I think this looks good. Thank you!

PR #253, opened 2017-02-24
DOMPurify testcases

g-k:
This is a first pass at adding testcases from the DOMPurify fixtures (specifically https://raw.githubusercontent.com/cure53/DOMPurify/master/test/fixtures/expect.js).

I'm not sure:

1. the conversion script here is correct (with respect to character encodings and how to handle multiple expected values)
1. how closely bleach's behavior matches DOMPurify

However, Cure53 has done a lot of XSS work (e.g. for Github) so there should be some interesting vectors here to test for in bleach too.

Note: this is based off https://github.com/mozilla/bleach/pull/251 so it'll need to be rebased, but as it stands a bunch of the new tests fail so this isn't ready to go in anyway.


g-k:
Closing for now.

I'll try to add these a few at a time so I can review them and we can have them passing right off the bat.

PR #256, opened 2017-03-03
Refactor allow_token

willkg:
This redoes the innards of allow_token so that it's a single loop across all the
attributes in a token rather than a bunch of little passes.

This has less looping, so theoretically it's more optimal, but I didn't spend
any time testing that theory.


willkg:
FYI: This PR should be cosmetic in regards to behavior. The end result should be exactly the same--this just redoes the looping so we're looping less to do it.

PR #257, opened 2017-03-03
Add cleaner

willkg:
Fixes #257.


willkg:
Good enough. We can fix issues as they come up.

PR #259, opened 2017-03-03
Implement ability to use Filters in cleaning

### LINKED ISSUE #58: A cleaning pipeline

willkg:
This lets users extend cleaning more easily by specifying an html5lib Filter.

Fixes #58.

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -115,9 +115,8 @@ class Cleaner(object):
 
     def __init__(self, tags=ALLOWED_TAGS, attributes=ALLOWED_ATTRIBUTES,
                  styles=ALLOWED_STYLES, protocols=ALLOWED_PROTOCOLS, strip=False,
-                 strip_comments=True):
-        """
-        :arg tags: whitelist of allowed tags; defaults to
+                 strip_comments=True, filters=None):
+        """:arg tags: whitelist of allowed tags; defaults to
             ``bleach.ALLOWED_TAGS``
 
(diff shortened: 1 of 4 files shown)


willkg:
Looks good.

Issue #181, opened 2015-12-03
Make linkify extensible 


defrex:
URLs and Email addresses are two types of things you might want to linkify, but they aren't the only things. For example, in #178 I extended it to linkify hashtags. Other things like Issues and PRs on GubHub, or any other application-specific thing one could think of, could all benefit from `bleach.linkify`.

An API could look something like this:

``` Python
import re
import bleach

hashtag_re = re.compile(r'#([\w\-_]+)', re.IGNORECASE | re.MULTILINE)


def hashtag_handler(re_match):
    tag = re_match.group(1)
    return {
        '_text': '#{}'.format(tag),
        'href': '/itemlist/?tag={}'.format(tag),
    }


bleach.linkify(input, extra_types=(
    (hashtag_re, hashtag_handler),
))
```

willkg:
I've thought about this and I see the utility, but it's not something I want to implement in Bleach.

However, I just rewrote linkify as an html5filter. That'll go out with Bleach 2.0. Using that, you can hang another filter off of that and implement whatever kind of additional linkifying you want to do.

Given that, I'm going to close this out.

psychomantys:
Theres a example link for this on bleach 2.0?

willkg:
@psychomantys Maybe these help?:

* https://bleach.readthedocs.io/en/latest/linkify.html#using-bleach-linkifier-linkifyfilter
* https://html5lib.readthedocs.io/en/latest/movingparts.html#filters
* reading the html5lib code for filters

psychomantys:
Uhum, i need to implement a filter on html5lib?

Is more easy to use a regex on the text before linkify it?

willkg:
Yes, you need to write a html5lib Filter. You can use regexes in them--that'd be fine. If you don't like that, you can call linkify and then run your regex on the string output. Lots of options!

psychomantys:
Thanks!! ^^

Issue #165, opened 2015-09-24
Escaped iframe HTML becomes broken iframe HTML with linkify


MadSpindel:
If you have escaped HTML of an iframe:

``` html
&lt;iframe width="480" height="360" src="https://www.kickstarter.com/projects/tomberryart/on-the-other-side-of-town/widget/video.html" frameborder="0" scrolling="no"&gt; &lt;/iframe&gt;
```

and runs it with bleach.linkify() it becomes real HTML (but broken):

``` html
<iframe class="linkify" height="360" https:="" on-the-other-side-of-town="" projects="" src="<a href=" target="_blank" tomberryart="" video.html"="" widget="" width="480" www.kickstarter.com=""><a class="linkify" href="https://www.kickstarter.com/projects...get/video.html" target="_blank">https://www.kickstarter.com/projects...get/video.html</a>" frameborder="0" scrolling="no"> </iframe>
```

It should not convert escaped HTML!

How I use bleach.linkify:

``` python
def shorten_url(attrs, new=False):
    """Shorten overly-long URLs in the text."""
    if not new:  # Only looking at newly-created links.
        return attrs
    # _text will be the same as the URL for new links.
    text = attrs['_text']
    if len(text) > 53:
        attrs['_text'] = text[0:36] + '...' + text[-14:]
    return attrs


def set_target(attrs, new=False):
    if not new:  # Only looking at newly-created links.
        return attrs
    attrs['target'] = '_blank'
    attrs['class'] = 'linkify'
    return attrs

DEFAULT_CALLBACKS = [set_target, shorten_url]

@register.filter(is_safe=True)
def linkify(value):
    return bleach.linkify(value, callbacks=DEFAULT_CALLBACKS, skip_pre=False, parse_email=False, tokenizer=HTMLTokenizer)
```

nikolas:
Interesting problem .. I'm not sure how to solve it. One way I can think of is adding a `skip_src` argument to `linkify`, similar to the `skip_code` argument I've proposed here: https://github.com/jsocol/bleach/pull/162

There might be a nicer way, though.

jsocol:
Linkify shouldn't be unescaping valid character references (e.g. `&lt;`) and _especially_ not meaningful entities. There's no need for more kwargs: this is a bug.

willkg:
This is fixed with the linkify rewrite.

Issue #153, opened 2015-05-12
Linkify strips href attribute from exsiting relative links


alexington:
The `bleach.linkify()` method removes the href attribute of existing links with relative paths.

Absolute path link:

```
>>> bleach.linkify('<a href="http://google.com">example</a>')
u'<a href="http://google.com" rel="nofollow">example</a>'
```

Relative path link:

```
>>> bleach.linkify('<a href="/relative/path/">example</a>')
u'<a>example</a>'
```

I'd like to avoid this, since I want to retain all existing links (including relative path links).  But I'm not able to figure out how to do that -- I'm probably just missing something obvious!  Thanks for any suggestions.

mauritsvanrees:
The same is happening now on PyPI.  The table of contents for packages is stripped of anchor links, which are also relative links.

This has been fixed in html5lib. From the changelog:

```
0.999999/1.0b7
Released on July 7, 2015

Fix #189: fix the sanitizer to allow relative URLs again (as it did prior to 0.9999/1.0b5).
```

willkg:
This is fixed with the linkify rewrite.

Issue #78, opened 2012-10-30
Error in bleach.linkify() using links enclosed by parens and a dot


barsch:
Using just brackets works fine:

``` python
>>> import bleach
>>> bleach.linkify('hello (http://www.mu.de/blah.html) world')
u'hello (<a href="http://www.mu.de/blah.html" rel="nofollow">http://www.mu.de/blah.html</a>) world'
```

adding a dot results into an incorrect link:

```
>>> bleach.linkify('hello (http://www.mu.de/blah.html). world')        
u'hello (<a href="http://www.mu.de/blah.html)" rel="nofollow">http://www.mu.de/blah.html)</a>. world'
```

willkg:
This is fixed with the linkify rewrite.

PR #260, opened 2017-03-04
Linkify

### LINKED ISSUE #46: Allow method chaining without rebuilding the tree
Right now, if you chain methods (e.g. `linkify(clean(text))`) the whole tree-building, which is expensive, has to happen twice.

I can see two ways of solving this. One is a total API rewrite, one is less complicated. I'm leaning toward the latter.

**Total rewrite:**

Recreate a `Bleach` class, but this time it _is_ the text you want to clean, e.g.:

``` python
from bleach import Bleach
text = Bleach(text)
print text.clean()
```

It could store state of what's already been done (like cleaning, linkifying, etc) and either return text (as above) or return itself to allow chaining, e.g.:

``` python
# Implement __unicode__ and __str__, seems nicer than above.
print Bleach(text).clean().linkify()
```

**Less-crazy idea:**

Instead of actual `unicode` objects, the top-level methods return a lazy object that _looks_ like a string, maybe even descends from `unicode`, but is smarter, stores the tree, and only renders itself if anything needs to treat it as a string.

Then the methods would be reworked to check for the tree before rebuilding it, and if found, they would just use the existing tree. `clean()` might need special handling--i.e. always go back to text first--because of the nature of what it does. But I generally say you should always call `clean()` first if you're chaining, anyway.

**Pros and cons**

The rewrite might make it easier to deal with the increasing complexity of `linkify()`. You could (again) subclass `Bleach` and override a number of things there. I don't love APIs where you have to subclass things. I'm also not sure how complex this would make the upgrade for consumers--it would be a major version change.

The less-crazy idea doesn't involve subclassing anything, ever. The API can stay the same and it can be a minor version change. I am guessing won't be _too_ difficult to write some sort of `LazyTreeStringObjectThing` class. It would be a good opportunity to do some refactoring and get some stuff out of `bleach/__init__.py`. (I wouldn't change import paths, just where the code lives.)

### LINKED ISSUE #130: bleach linkify unescapes escaped HTML sometimes
@jsocol, thanks for your library, we use it for years :+1: 

Recently we discovered quite a strange behaviour `linkify` exposes in one scenario. Please, take a look on my example below. As you can see, sometimes it unescapes a correctly escaped tag:

``` python

In [1]: import bleach

In [2]: bleach.__version__
Out[2]: u'1.4'

In [3]: bleach.linkify('&lt;br&gt;')  # everything is okay
Out[3]: u'&lt;br&gt;'

In [4]: bleach.linkify('&lt;br&gt; http://example.com')  # why <br> is unescaped?
Out[4]: u'<br> <a href="http://example.com" rel="nofollow">http://example.com</a>'

In [5]: bleach.linkify('&lt;br&gt; <br> http://example.com')  # ... and here it's not again ;-)
Out[5]: u'&lt;br&gt; <br> <a href="http://example.com" rel="nofollow">http://example.com</a>'
```

Frankly, I have no idea how it should be fixed. Thanks for any input.

### LINKED ISSUE #200: linkify hanging for this string:
With fresh install of bleach (1.4.3) in a new virtualenv:

```
>>> import bleach
>>> bleach.linkify("an@email.com<mailto:an@email.com>", parse_email=True)
... hangs ...
```

(found because I have it linkifying emails - and this came through in someone's sig)

willkg:
This is a rewrite of linkify and all that stuff.

Fixes #46.

Fixes #200.

Fixes #130.

### DIFF
--- a/bleach/linkifier.py
+++ b/bleach/linkifier.py
@@ -254,7 +254,6 @@ class LinkifyFilter(Filter):
 
                     else:
                         # Add the "a" tag!
-
                         if prefix:
                             new_tokens.append(
                                 {u'type': u'Characters', u'data': prefix}
@@ -289,6 +288,49 @@ class LinkifyFilter(Filter):
 
             yield token
 
+    def handle_a_tag(self, token_buffer):
+        """Handle the "a" tag
+
+        This could adjust the link or drop it altogether depending on what the
+        callbacks return.
+
+        This yields the new set of tokens.
+
+        """
+        a_token = token_buffer[0]
+        if a_token['data']:
+            attrs = a_token['data']
+        else:
+            attrs = {}
+        text = self.extract_character_data(token_buffer)
+        attrs['_text'] = text
+
+        attrs = self.apply_callbacks(attrs, False)
+
+        if attrs is None:
+            # We're dropping the "a" tag and everything else and replacing
+            # it with character data. So emit that token.
+            yield {'type': 'Characters', 'data': text}
+
+        else:
+            new_text = attrs.pop('_text', '')
+            a_token['data'] = alphabetize_attributes(attrs)
+
+            if text == new_text:
+                # The callbacks didn't change the text, so we yield the new "a"
+                # token, then whatever else was there, then the end "a" token
+                yield a_token
+                for mem in token_buffer[1:]:
+                    yield mem
+
+            else:
+                # If the callbacks changed the text, then we're going to drop
+                # all the tokens between the start and end "a" tags and replace
+                # it with the new text
+                yield a_token
+                yield {'type': 'Characters', 'data': force_unicode(new_text)}
+                yield token_buffer[-1]
+
     def __iter__(self):
         in_a = False
         in_pre = False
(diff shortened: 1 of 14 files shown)


bleach/linkifier.py
```diff
@@ -0,0 +1,356 @@
...
+                    in_a = True
+                    token_buffer.append(token)
+
+                    # We buffer the start tag, so we don't want to yield it,
+                    # yet
+                    continue
+
+            elif in_pre:
+                # NOTE(willkg): We put this clause here since in_a and
+                # switching in and out of is_a takes precedence.
```
willkg: Typo--the `is_a` should be an `in_a`.

bleach/utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def alphabetize_attributes(attrs):
+    """Takes a dict of attributes (or None) and returns them alphabetized"""
+    if not attrs:
+        return attrs
+
+    return OrderedDict(
+        [(k, v) for k, v in sorted(attrs.items(), key=_attr_key)]
+    )
```
willkg: I should add some tests for this.

willkg:
This allows people to do use cases specified in #181, so I think this covers that, too.

willkg:
This also covers #165.

willkg:
This covers #153.

willkg:
This covers #78, too.

bleach/linkifier.py
```diff
@@ -0,0 +1,387 @@
...
+                        new_tokens.append({u'type': u'Characters', u'data': text[end:]})
+
+                    for new_token in new_tokens:
+                        yield new_token
+
+                    continue
+
+            yield token
+
+    def strip_non_url_bits(self, fragment):
```
willkg: One nice thing about pulling all this functionality into this single method is that someone can override/extend it with a `SuperLinkifyDoWhatIWantFilter`.

willkg:
Seems good to me. Landing this!

kousu:
I really appreciate you taking a look at #130 and getting this out to PyPI! From my perspective, it was fixed almost immediately. That's amazing! :clap:

Issue #223, opened 2016-10-31
TLD list is not comprehensive and is outdated affecting linking


willkg:
bleach has a hard-coded set of TLDs that it uses for the `url_re` which is used for linkifying.

The problem with this is that it's not updated and it's not exposed to devs so they have no ability to adjust it.

PR #159 covered adding some infrastructure to provide an updated list along with infrastructure to update the list for your local copy of bleach.

However, I think that's not a road I want to go down.

I've been wrestling with a few aspects of that PR and general direction:

1. If we land this or something like it, how do we maintain a list of TLDs going forward? Does that require us to do regular releases? Does that require us to do releases that are solely a TLD update? 
2. Are there security issues around TLDs that would be better served by separating maintenance of a list into a separate project?
3. This PR provides a command to update the TLD list, but we have no insight as to whether that operation is successful. It might be successful now, but what if it fails tomorrow? Will a user know that it's bad? Is there a way for a user to verify it? What's our responsibility for making sure it always works correctly?

I think requests has a similar situation where they used to have CA bundles in the requests package and then split that out into a separate project called certifi. Then certifi updates can come out independently of requests and it provides a nice API for people to use their own CA bundles.

I think treating this the same way is worth looking into. I want to take a look around to see what the current state of TLD-related Python libraries is and maybe we should use one of them.

This issue covers researching other TLD-related Python libraries--can any of them provide a certifi-like function for bleach? If so, let's do that. If not, then we'll probably have to start one ourselves.

EmilStenstrom:
Alternatives:
* tld: https://pypi.python.org/pypi/tld
* tldextract: https://pypi.python.org/pypi/tldextract
* publicsuffix: https://pypi.python.org/pypi/publicsuffix/
* publicsuffixlist: https://pypi.python.org/pypi/publicsuffixlist/
* dnspy: https://pypi.python.org/pypi/dnspy/

None of them release a new version every time there is an update to the tld list. If we want that we need to build it. Instead they either require that the user supplies an updated list of that you run an update script before you get started. 

Notes:
* tldextract handles the [differences between ICANN domains and Private domains](https://github.com/john-kurkowski/tldextract/issues/19) properly (exclude the private ones)
* dnspy hasn't been updated since 2013.
* dnspy does not support python 3.
* All of them have permissive licences (MPL, BSD, or MIT).

I think tldextract looks like the best option if we should use an external library.

Resources:
[...]

Issue #55, opened 2012-02-17
Refactor, clean up __init__


jsocol:
Right now, all the interesting code in Bleach is in `__init__.py`, which is convenient but not a great practice. I'd like to shuffle things around a little and clean up that file.

willkg:
PR #261 covered a rewrite. It's much nicer now (in my opinion). As such, closing this out.

PR #261, opened 2017-03-06
Restructure linkify; clean up __init__; update docs

willkg:
This is a ton of changes all in one commit. Sorry.

* Restructures linkify so it mirrors clean. This has the nicety in that
  the two are parallel and can be used the same.
* Rework how url_re and email_re work so that it's possible to override
  them and/or provide your own list of allowed procotols and TLDs
* Overhaul the docs including converting linkify examples to doctest
* Update CHANGES

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,7 +17,7 @@ Version 2.0 (in development)
 
 * ``bleach.clean`` and friends were rewritten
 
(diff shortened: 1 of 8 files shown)


willkg:
This doesn't fix #223, but it alleviates it and makes it easier for users to do their own TLD thing.

This fixes #55.

willkg:
Seems fine.

PR #263, opened 2017-03-07
Rework attributes value and filters

### LINKED ISSUE #126: improved callable filter support
this is perhaps a bad title.  i'll explain the problem:

as mentioned in an earlier ticket, i needed to support the  "data-*" namespace on attributes.  the suggested route is callable filters.

the two main problems are as follows
- `*` doesn't accept callable filters.  based on the current setup, it requires an iterable.
- specific tags can have a callable filter OR a list of attributes.  not both.

both of these are invalid:

```
macro_attributes_ok = {
    '*': allowable_macro_attributes,
    'a' : ['href', 'title', 'target', 'rel', allowable_macro_attributes],
}
```

the callable_filters, as-is, must be implemented with a dedicated function per TAG.  This gets messy as a lot of code must be written.

```
macro_attributes_ok = {
    'a' : allowable_attributes_a,
    'div' : allowable_attributes_div,
    'span' : allowable_attributes_span,
}
```

approaches that could be nicer / easier to deal with:

• allow for callables to be within a list of text attributes

```
macro_attributes_ok = {
    'a' : [ 'href', 'link', allowable_attributes_setA ],
    'div' : [ 'class', 'style', allowable_attributes_setA ],
}
```

• pass the tag's name in as a kwarg to the callable.  this would allow callables to reflect the current tag -- and be backwards compatible.  

```
def filter_src(name, value, tag=None):
    if tag != 'img':
         return False
    if name in ('alt', 'height', 'width'):
        return True
```
-  the `attributes` kwarg to clean could be a callable itself; one that expects a triplet of data
  
     def validate_attribute( tag, attribute, value ):
           return True
  
    clean( text, attributes=validate_attribute)

willkg:
This reworks how attributes argument works. Callables now take three arguments:
tag, attribute name and attribute value. Callables can be passed in as the
attributes argument value or as a value for any of the tags in the dict.

This also reworks the implementation so the complexity of the different shapes
is shuffled away out of ``allow_token`` which simplifies it a bit.

Fixes #126.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -25,6 +25,12 @@ Version 2.0 (in development)
   Amongst other things, this version will add end tags even if the tag in
   question is to be escaped.
 
+* ``bleach.clean`` and friends attribute callables now take three arguments:
+  tag, attribute name and attribute value. Previously they only took attribute
+  name and attribute value.
+
+  All attribute callables will need to be updated.
+
 * ``bleach.linkify`` was rewritten
 
   ``linkify`` was reimplemented as an html5lib Filter. As such, it no longer
@@ -52,6 +58,8 @@ Version 2.0 (in development)
   don't then html5lib will raise an assertion error that the value is not
   unicode.
 
+  All linkify filters will need to be updated.
+
 **Changes**
 
 * Supports Python 3.6.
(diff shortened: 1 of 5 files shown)


willkg:
Looks fine.

PR #273, opened 2017-03-20
Fixed KeyError exception on tags without href attr

defsen:


bleach/callbacks.py
```diff
@@ -4,7 +4,11 @@
 
 def nofollow(attrs, new=False):
     href_key = (None, u'href')
-    if href_key not in attrs or attrs[href_key].startswith(u'mailto:'):
+
+    if href_key not in attrs:
+        return attrs
+
+    if attrs[href_key].startswith(u'mailto:'):
```
willkg: Why'd you make this change? It doesn't seem like it changes the behavior at all.

bleach/callbacks.py
```diff
@@ -18,6 +22,10 @@ def nofollow(attrs, new=False):
 
 def target_blank(attrs, new=False):
     href_key = (None, u'href')
+
+    if href_key not in attrs:
+        return attrs
```
willkg: Good catch!

willkg: Looks like `target_blank` has no tests. Can you add some?

bleach/callbacks.py
defsen: Oops, missed it.

bleach/callbacks.py
defsen: Of course.

willkg: Mmm... It looks like neither of these callbacks have any tests. That's terrible. Sorry about that. I'll throw some together after landing this.

willkg:
Thank you for fixing this!

PR #274, opened 2017-03-20
Add tests for linkify callbacks

willkg:
* adds tests for linkify callbacks
* fix linkify calls to always specify "callbacks" argument name


willkg:
When reviewing #273, I discovered there were no tests for these callbacks. This adds some.

Issue #277, opened 2017-04-06
Please document difference between BleachSanitizerFilter and html5lib’s sanitizer.Filter


stucox:
It’s not clear what the difference is between using Bleach’s sanitizer and the one which ships with html5lib.

`BleachSanitizerFilter` inherits from `html5lib.filters.sanitizer.Filter` and overrides the `sanitize_token()`, `allow_token()` and `sanitize_css()` methods, but even reading the code isn’t entirely obvious what the functional difference is.

It does have a slightly different API and different defaults — are they the only differences? Or given equivalent config (tags, attributes, etc), would they give the same output?

If not, in what way(s) is Bleach’s version better?

willkg:
Sure! Actually, one of the things I did for Bleach 2.0 was restructure Bleach's sanitizer so it was easier to compare the two. Sounds like I didn't do a good enough job on that. Sorry!

Bleach's sanitizer allows for a map for `ALLOWED_ATTRIBUTES` giving you a lot more control over sanitizing attributes: you can sanitize attributes for specific tags, you can sanitize based on value, etc.

Bleach's sanitizer always alphabetizes attributes, but uses an alphabetizer that works with namespaces--the html5lib one is broken in that regard.

Bleach's sanitizer always quotes attribute values because that's the safe thing to do. The html5lib one makes that configurable. In this case, Bleach doesn't make something configurable that isn't safe.

Bleach's sanitizer has a very restricted set of `ALLOWED_PROTOCOLS` by default. html5lib has a much more expansive one that I claim is less safe.

`html5lib.filters.sanitizer.Filter`'s `sanitize_css` is broken and doesn't work.

There are probably some other differences, but they're more minor.

Hope that helps!

stucox:
This is great, thanks. I’ve raised PR #278 to discuss adding this to the README.

Will close this off; further discussion can happen on the PR.

PR #278, opened 2017-04-13
Add comparison of Bleach vs html5lib to README

stucox:
Used notes provided by @willkg on #277 to add a short description of why Bleach may be preferable to html5lib’s built-in sanitizer.

Let me know if you think this should live somewhere else.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,6 +89,19 @@ The simplest way to use Bleach is:
     u'an <a href="http://example.com" rel="nofollow">http://example.com</a> url
 
 
+Bleach vs html5lib
+==================
+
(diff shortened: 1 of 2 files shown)


willkg:
Thank you for writing this up! Since you asked your question in #277, I've been looking around to see if it's a question that other potential users might be asking and haven't seen any evidence of this, but it's interesting to have written down for future Bleach maintainers/developers. There isn't a list anywhere else like that except my head (and now a comment in #277).

I tend to want to keep the README short and to-the-point and fairly static over time. I think a bunch of "bleach vs. xyz" comparisons are hard to do in a way that they maintain their truth over time and I'm not sure this information is pressing for enough users to warrant being in the README.

Given that, I'd put it somewhere else in the docs instead. Maybe in the Goals chapter at the end since the important differences between the two sanitizing implementations come from a difference in mission?

Does that sound like a good idea?

I do appreciate your efforts! Thank you!

stucox:
Yes, certainly — I’ll update my PR.

willkg:
Thank you!

PR #284, opened 2017-05-02
Fix html5lib requirements

indivisible:
Mark package compatible with html5lib versions >=1.0b9.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ tests_require = [
 install_requires = [
     'six',
     # >= 8 9s because of breaking API change
-    'html5lib>=0.99999999',
+    'html5lib>=0.99999999,>=1.0b9',
 ]
 
 
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,8 @@ tests_require = [
 install_requires = [
     'six',
     # >= 8 9s because of breaking API change
-    'html5lib>=0.99999999,>=1.0b9',
+    # the 'pre' suffix is needed for supporting '1.0b*' versions
+    'html5lib>=0.99999999pre,!=1.0b1,!=1.0b2,!=1.0b3,!=1.0b4,!=1.0b5,!=1.0b6,!=1.0b7,!=1.0b8',
 ]
 
 
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py{27,33,34,35,36}-html5lib{99999999,999999999,1.0b9,1.0b10},pypy-html5lib99999999
+envlist = py{27,33,34,35,36}-html5lib{99999999,999999999,10b9,10b10},pypy-html5lib99999999
 
 [testenv]
 basepython =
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -16,7 +16,7 @@
 install_requires = [
     'six',
     # >= 8 9s because of breaking API change
-    'html5lib>=0.99999999',
+    'html5lib>=0.99999999,>=1.0b9',
```
willkg: How does this work? Seems like the second `>=` overrides the first since a version number can't satisfy both clauses.

indivisible: I'm not sure why it works, but it does.

If I do an install, pip will pull 1.0b10. If I install html5lib== the 8-nines version, pip check will say it's alright. If I install b8 or a 7-nines version pip check complains of a broken dependency.

willkg: This felt wrong, so I spent some time looking for the specification that dictates how dependency versions are specified.

https://www.python.org/dev/peps/pep-0440/#version-specifiers

According to that, the `,` denotes an AND between version clauses. Thus this change says "use html5lib >= 0.99999999 AND >= 1.0b9" which doesn't make any sense.

Seems like the more appropriate way to do this is something like:

```
html5lib>=0.99999999,!=1.0b1,!=1.0b2,!=1.0b3,!=1.0b4,!=1.0b5,!=1.0b6,!=1.0b7,!=1.0b8
```

That says "html5lib >= 0.99999999 and, oh, by the way, don't allow this set of known-bad versions".

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py{27,33,34,35,36}-html5lib{99999999,999999999},pypy-html5lib99999999
+envlist = py{27,33,34,35,36}-html5lib{99999999,999999999,1.0b9,1.0b10},pypy-html5lib99999999
```
willkg: These additional two `html5lib` versions need to get things added to the `deps` clause below. Something like this:

```
    html5lib1.0b9: html5lib==1.0b9
    html5lib1.0b10: html5lib==1.0b10
```

Also, I'm not sure if periods are allowed in environment names--so we should look into that.

willkg:
I really appreciate you tackling this! I think if you fix those two things, this is good to go. Thank you!

setup.py
indivisible: That might be the best way. The funny thing is that they managed to choose a pre-release version scheme for the saner versions, while the all-nines versions get interpreted as proper release versions, and so don't get selected normally.

setup.py
```diff
@@ -16,7 +16,8 @@
 install_requires = [
     'six',
     # >= 8 9s because of breaking API change
-    'html5lib>=0.99999999',
+    # the 'pre' suffix is needed for supporting '1.0b*' versions
```
willkg: I don't understand this comment. What do you mean by this?

indivisible: As in 0.99999999**pre**. It's there to have pip treat it as a dependency on a prerelease version. Otherwise it won't select prerelease versions like 1.0b9, since it sees "0.99999999" and thinks it's a proper release. [See this for its effects](https://gist.github.com/indivisible/5f70480df776f48907218006278da5af).

willkg: Ahh... interesting. Do we know if that's dependent on pip versions or does it work with old pip versions, too?

indivisible: I develop on py3.6, pip 9.0.1. Tested to work on py3.5 with pip 7.0.0.

Looking at the included libraries it will not work with pip 6.0.

willkg:
Thank you!

PR #286, opened 2017-05-19
docs: Add XSS gadget safe escaping as a non-goal

g-k:
Per IRC discussion. Please modify as necessary.


willkg:
This is great! Thank you!

Issue #62, opened 2012-04-20
AssertionError: Node has no children


lestatcheb:
If I do

```
value = bleach.clean('<p><iframe frameborder="0" height="315" src="http://www.youtube.com/embed/6ydeY0tTtF4" width="560"></iframe></p>', strip=True, tags=[], attributes=[])
```

it's ok that it print assertionError?

```
2012-04-20 15:23:27,885 ERROR   /usr/lib/python2.7/site-packages/bleach/__init__.py:328 HTML: AssertionError('Node has no children',)
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/bleach/__init__.py", line 326, in _render
    return force_unicode(_serialize(tree))
  File "/usr/lib/python2.7/site-packages/bleach/__init__.py", line 341, in _serialize
    return serializer.render(stream)
  File "/usr/lib/python2.7/site-packages/html5lib/serializer/htmlserializer.py", line 302, in render
    return u"".join(list(self.serialize(treewalker)))
  File "/usr/lib/python2.7/site-packages/html5lib/serializer/htmlserializer.py", line 192, in serialize
    for token in treewalker:
  File "/usr/lib/python2.7/site-packages/html5lib/treewalkers/_base.py", line 154, in __iter__
    firstChild = self.getFirstChild(currentNode)
  File "/usr/lib/python2.7/site-packages/html5lib/treewalkers/simpletree.py", line 58, in getFirstChild
    assert node.hasContent(), "Node has no children"
AssertionError: Node has no children
```

value is u'', but in python shell printed it assertion error.

Maybe don't print it?

Thanks.

Issue #60, opened 2012-03-28
Some email addresses are messed up with parse_email=False


ThiefMaster:
```
>>> linkify('doesntmatter@foo-bar.de')
u'doesntmatter@foo<a href="http://-bar.de" rel="nofollow">-bar.de</a>'
```

willkg:
I'm going to drop this from the milestone. I think it needs more thinking and I'm not going to get to it for v2.0.

willkg:
This is a problem, but this isn't something we're going to fix. Sorry!

Issue #38, opened 2011-08-17
linkify doesn't like ports


jsocol:
`http://example.com:9000/foo` should get linkified as a whole, but it stops at `.com`.

PR #293, opened 2017-07-27
143 entities

### LINKED ISSUE #143: HTML entities being unescaped
I don't know if this is related to #130 , but I have an issue with this with non-breaking space being unescaped:

```
import bleach
bleach.clean('&nbsp;')
# results in -> u'\xa0'
```

Bleach version 1.4.1
HTML5lib version 0.999

willkg:
This fixes #143.


tests/test_security.py
```diff
@@ -181,6 +192,6 @@ def test_regression_manually():
     """Regression tests for clean so we can see if there are issues"""
     # NOTE(willkg): Have to do this one by hand because of the \r
     s = """<IMG SRC="jav&#x0D;ascript:alert(<WBR>'XSS');">"""
-    expected = """&lt;img src="jav\rascript:alert(&amp;lt;WBR&amp;gt;'XSS');"&gt;"""
+    expected = """&lt;img src="jav&amp;#x0D;ascript:alert(&lt;WBR&gt;'XSS');"&gt;"""
```
jwhitlock: Now that the ``\r`` is gone, does this note still apply?

willkg: No, it doesn't! Plus it means we can fold it into the other tests. Good catch!

jvanasco:
Initial thoughts:

Preface- This is long.  *Please don't hate me!*  I think the PR is great, but diving deeper into it with some integrated test cases based on usage, it seems bleach was broken in other unreported ways and this introduces some larger/undesirable regressions on a "usage" scale.  I was trying to think about usage in terms of end-users and use-cases, not just unit tests.  

Going through all this, I think bleach performed a few functions in a 'not great' technical manner, but few people cared because the end-result was fine.  It seems like this PR takes a technically correct and elegant approach to solving some issues, but risks breaking current usage patterns because the inelegant approach 'just worked'.

1- The clarification on HTML attributes looks good. It's not necessary due to the fixes PR, but a good idea (our issue was that we put "attribute-safe" data into bleach, but it was decoded then "cleaned" into an unsafe format.)

2- I'm worried by the loss of idempotent transforms. It doesn't affect us, but this has the potential to really break stuff.  I know it may not be avoidable.   IMHO this would need to be a 2.1 not 2.0.1 release with a big warning. 

3- The new encoding behavior will break installs where html entities are entered, which includes many wyswig markdown editors.

consider this high-level use-case that is largely oblivious to what is going on:

	comment = Comment(user_id)
	comment.content = bleach.clean('&bull;')
	session.save(comment)  # will store to utf-8
    
the old behavior would decode the `&bull;` to the unicode point `u'\u2022'`; then the database would store it as utf-8 behind the scenes. 

	bleach.clean('&bull;').encode('utf-8')
	'\xe2\x80\xa2'

when this gets rendered out, a bullet still renders on the HTML page.  it's no longer a `&bull;` but the unicode character... however viewers will still see a bulletpoint.

under the new transformations, `&bull;` will be appear where a bulletpoint is expected (as `u'&amp;bull;'` is stored in the database)

This is a GIANT change and likely to adversely affect many.

4- A tiny breaking change is how bleach handles the "special" `cgi.escape` entities if entered in the numeric form.  previously these were transcoded from numeric to text.

	input = """&#38;&#60;&#62;"""
	1.4.3 and 2.0.0:
		bleach.clean("""&#38;&#60;&#62;""")
		u'&amp;&lt;&gt;'
	pr#293
		u'&amp;#38;&amp;#60;&amp;#62;'

I doubt anyone relied on this or realized it, but I did want to mention it.

5- The clean and linkify commands on query strings don't create the same output, still break links, and don't necessarily work as expected:

	plaintext_url = 'http://example.com?active=true&current=true'
	bleach.clean(plaintext_url)
	u'http://example.com?active=true&amp;current=true'
	
	html = '<a href="http://example.com?active=true&current=true" rel="nofollow">http://example.com?active=true&current=true</a>'
	bleach.clean(html)
	u'<a href="http://example.com?active=true&amp;current=true">http://example.com?active=true&amp;current=true</a>'

	bleach.linkify(plaintext_url)
	u'<a href="http://example.com?active=true\xa4t=true" rel="nofollow">http://example.com?active=true\xa4t=true</a>'

So a few things on the above.

`clean`

if we're just dealing with a plaintext_url, the encoding of the `&amp` upgrades this into an "html display of a url".  this is a bit of a conceptual shift and may break implementations that don't expect plaintext url occurrences to be upgrade to html.  For example, we run bleach before regexing input for urls (we track the domains/kwargs to identify spammers).  We'll need to parse the output as html first now.

if we're dealing with html input, this breaks the url in the href when there are more than one param(!).  this appears to be previously broken back to at least 1.4.3 , so it's not a new problem - but it is a problem (I'll create a new issue for this)

expected outputs would need to preserve the URL in attributes:

	u'<a href="http://example.com?active=true&current=true">http://example.com?active=true&current=true</a>'
	u'<a href="http://example.com?active=true&current=true">http://example.com?active=true&amp;current=true</a>'


Personally, I would prefer to see the 'visible' representation of a url to NOT have ampersands quoted.  Aside from problems with tracking urls, it breaks a usage pattern where cleaned text linkified :

	bleach.linkify(bleach.clean(input)))
	
For example, we store 'cleaned' comments, but also linkify them... 

`linkify`

This is still broken - perhaps it uses a different parser?

Look carefully at the transformation of the querystring args on linkify 

	active=true&current=true
	active=true\xa4t=true
	
*i did not get to dive into the tokenizer yet* but it superficially appears the parser sees the non-entity usage of `&` and is consuming the first 6 chars as if it were an entity (so `&current=` becomes `\xa4t=true`).  This is just a superficial guess as there are a few valid entities with 7 chars (`&1234567;`).

In any event, with the above link an "expected" output of clean would be to not encode the ampersand in the attributes:

	u'<a href="http://example.com?active=true&current=true">http://example.com?active=true&current=true</a>'
	u'<a href="http://example.com?active=true&current=true">http://example.com?active=true&amp;current=true</a>'

willkg:
@jvanasco  I appreciate you taking the time to think through this!

My thoughts below:

> 1- The clarification on HTML attributes looks good. It's not necessary due to the fixes PR, but a good idea (our issue was that we put "attribute-safe" data into bleach, but it was decoded then "cleaned" into an unsafe format.)

`bleach.clean()` output is still **not** safe for HTML attributes. You need to use your templating engine to escape the output of `bleach.clean()`. This PR doesn't fix that hence the additional documentation.

> 2- I'm worried by the loss of idempotent transforms. It doesn't affect us, but this has the potential to really break stuff. I know it may not be avoidable. IMHO this would need to be a 2.1 not 2.0.1 release with a big warning.

These changes are targeting a 2.1 release because it includes backwards-incompatible changes. That's covered in the changes in the `CHANGES` file. I think that covers your concerns.

Regarding idempotency, I'm not sure how sanitizing could ever be idempotent because of what's it's doing. There are notes in the first commit message about that.

If you think this is a necessary property for `bleach.clean`, I'm game for reading a compelling argument, but note that similar functions in other places aren't idempotent (escape, quote, and so on).

`bleach.linkify` is different. It could be idempotent since it's adding links to things.

> 3- The new encoding behavior will break installs where html entities are entered, which includes many wyswig markdown editors. ...

It seems like the use case here is that users use a wysiwig editor, type stuff in, and the editor adds additional things like bullet points using character entities. Then the system which can't differentiate between what the user wrote and what the editor added, runs all that text through `bleach.clean()` which will convert things like `&bull;` to `&amp;bull;`.

I'd love to know more about this use case. None of the major projects that use Bleach do this as far as I know.

My gut feeling is that what they're doing isn't a supported use case for Bleach and they're doing things to get around what Bleach does. That's fine, and you're right--these changes will probably break that. I'm not sure I think that's a bad thing or that it should stop us from landing this PR as is.

Do you have examples of systems like this so I can look at them?

Are there ways to fix the behavior we're fixing in this PR and also make that use case easier? It might be possible to consume entities specified in an allow list and leave the rest, but I think that creates new issues since `consumeEntity` can get called in a bunch of different contexts.

> 4- A tiny breaking change is how bleach handles the "special" cgi.escape entities if entered in the numeric form. previously these were transcoded from numeric to text.

This is covered in several of the tests.

Bleach is designed to take user-provided input and sanitize it for HTML output. If a user literally types in `&#38;`, are they expecting that to be a whitespace character when rendered in a webpage? That seems really curious to me.

What's the use case where `&#38;` ends up in the text and the expected behavior is that it comes out as whitespace?

> 5- The clean and linkify commands on query strings don't create the same output, still break links, and don't necessarily work as expected:

Your first two examples are actually correct behavior because the output of `bleach.clean()` is intended for HTML contexts. This covers both plain text as well as HTML attribute values. I think the previous behavior was incorrect.

However, `linkify` looks wrong. It's probably using the wrong parser. That should get fixed in this PR. I'll look at that either today or Monday.

I'm interested in seeing examples of wysiwig editors that use Bleach in the way you describe. It helps to have a list of actual users so I can run things by people before landing changes.

Thank you again for looking through it!

jvanasco:
Out of order...

> What's the use case where &#38; ends up in the text and the expected behavior is that it comes out as whitespace?

This is just a fringe case for the changelog. I don't know if anyone would rely on it.

Under previous versions: `&#38;` is transcoded to `&amp;`, which is then left as-is. The transcoding only applies to the 3 special characters in `cgi.escape`/`html.escape` (`<`, `>`, `&`), which go from numeric encoding to the text encoding.

> If you think this is a necessary property for bleach.clean, I'm game for reading a compelling argument, but note that similar functions in other places aren't idempotent (escape, quote, and so on).

I don't think it's necessary, just that it needs a clear warning sign that people will be sure to see.  

> I'm interested in seeing examples of wysiwig editors that use Bleach in the way you describe.

I don't know any wysiwig editors that are integrated with bleach.  my general point is that wysiwig editors tend to generate html entity defs, and that capability is broken under this PR.  We pull in some html entities via a wysiwg editor *and* consuming some JSON endpoints for data.

The rest of your comments all all somewhat related, and I need to jump deeper into the parser and functions to try some ideas out, but...

There are two concerns with the encoding of `&`:  html entities, and urls in content.

In terms of html entities:

I wonder if "safe" html entities need to be encoded as it means HTML entities being treated differently than tags, and like tags HTML entities are well-formed html.  The security concerns in the test suite are based on the dangers of escaping or omitting entities -- not having encoded ones.  By encoding & in this matter this manner, bleach is making every HTML fragment "unsafe" and basically generates HTML fragments to mimic the "plaintext" view.  How could someone properly allow an HTML entity under this system?  They're all effectively banned.

Encoding entities also has a fundamental shift in functionality -- it supposes that the input is plaintext instead of an html fragment, and the output is HTML that looks like the input.  

There are a few ways entities could be allowed.

potentially all entities could be left as-is (I'm not sure about the safety of this.  i believe it is safe, though it seems like the whitespace control ones may be unsafe in some javascript engines).  

python's stdlib tracks all the 'named' entities and their codepoints `six.moves.html_entities import name2codepoint`, and those should definitely be safe for rendering as-is (as the text or numeric input).  I played around with trying to wrap `bleach.clean` through a middleware like function to transcode these, but that needs to ignore the 'unsafe' entities [`('gt', 'lt', 'quot', 'amp')`  a single quote is `&#39;` but does not have a "name"]

In terms of urls in content:

i understand that `bleach` is merely guaranteeing content for an html context, but "upgrading" `&` to `&amp;` in a url  isn't making a string safer and (as stated above) breaks the utility of the library since `linkify` can not be applied to content that has been run through `bleach`.  

your design choice means developer must save the unsanitized content instead of bleached content, so link rendering can be controlled via `bleach(linkify(unsanitized))` .  saving the bleached content to later control link rendering via `linkify(bleached)` is **guaranteed** to break links because of the `&` to `&amp;` conversion.

willkg:
I think there's some confusion.

> i understand that bleach is merely guaranteeing content for an html context, but "upgrading" & to &amp; in a url isn't making a string safer and (as stated above) breaks the utility of the library since linkify can not be applied to content that has been run through bleach.

Earlier, you pointed out linkify is broken and I agreed and will look into it. So let's leave that out for now and just talk about `bleach.clean`.

Bleach 2.0 and previous versions convert `&` to `&amp;`:

```
> python
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
>>> import bleach
>>> bleach.clean('http://example.com/?this=that&foo=bar')
u'http://example.com/?this=that&amp;foo=bar'
>>> bleach.clean('<a href="http://example.com/?this=that&foo=bar">some link</a>')
u'<a href="http://example.com/?this=that&amp;foo=bar">some link</a>'
```

That doesn't change with this PR. What does change is this case:

```
> python
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
>>> bleach.clean('http://example.com/?this=that&current=bar')
u'http://example.com/?this=that\xa4t=bar'
```

In Bleach 2.0 and previous, `&curren` is interpreted as a character entity and changed to its unicode value. The change breaks the url.

With this change, those three examples become this:

```
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
>>> import bleach
>>> bleach.clean('http://example.com/?this=that&foo=bar')
u'http://example.com/?this=that&amp;foo=bar'
>>> bleach.clean('<a href="http://example.com/?this=that&foo=bar">some link</a>')
u'<a href="http://example.com/?this=that&amp;foo=bar">some link</a>'
>>> bleach.clean('http://example.com/?this=that&current=bar')
u'http://example.com/?this=that&amp;current=bar'
```

Those links look right and work fine in a browser.

Maybe I don't understand what you mean by urls broken by this change. Can you show me some examples of urls broken by `bleach.clean` changes in this PR?

jvanasco:
I'm incredibly sorry this isn't coming across correctly in text, as I'm trying to tie together a few concepts and as I noted above and the older versions of bleach are broken.  

Let me try and rephrase this differently.  The core concept is "The breaking change this PR introduces is the encoding of all HTML entities."  That breaks how input is visually displayed.  This problem also dovetails with some existing undocumented issues, which includes the URL (as I will explain in a moment)

Previous versions of bleach handled HTML entities incorrectly- they were decoded to unicode and then rendered as-is.  

    >>> print bleach.clean("&bull;")
    •
    >>> print bleach.clean("&bull;").__repr__()
    u'\u2022'

As this PR encodes the `&` , the new output is the following:

    >>> print bleach.clean("&bull;")
    &amp;bull;

This would render on a page as `&amp;bull;` and not the bullet character `•`.    This is the breaking change.

The related issue with URLs breaking should probably be a new issue as it is broken with existing code too.  I'll create a new ticket for that discussion and cross-reference.

jvanasco:
I added #294 to cover "Fragile Context" urls. 

After a lot of reading, I was wrong - most browsers will decode the `&amp;` and request `&`.  I'm sorry for being an old grouch on this.  

As you noted, the current behavior, and this PR, will generate the same URL for `bleach.linkify(bleach.clean(url))`.

willkg:
I mulled over this over the weekend. I'm coming around to the "it's not a good idea to convert all `&` to `&amp;`" thinking. I want to look into that more.

For now, I'm going to:

1. extract the documentation fix about using `bleach.clean` in HTML attributes out of this PR and land that in a separate PR
2. close this PR
3. go back to the drawing board and try some other things to see if there are behaviors that are sane, but less of a big change
4. open a new PR with whatever comes of that

Thank you!

jvanasco:
FWIW, I think you could achieve that with this flow:

* unescape html entities to unicode codepoints
* parse (perhaps using this PR)
* encode the unicode codepoints to html entities

PR #296, opened 2017-07-31
Clarify the dangers of using bleach.clean() output in html attributes

willkg:
This clarifies the dangers usage of `bleach.clean` output in HTML attributes with examples and some other commentary.


willkg:
This is extracted from PR #143, but doesn't rely on the rest of the changes in that PR. It's been looked over a couple of times in that PR, so I'm going to merge it.

Issue #292, opened 2017-07-19
bleach.linkify(None) returns u'None'


overfl0:
It looks like linkify is doing a repr(argument) somewhere in its code because I can pass it some other type arguments, like `int` (or even `type(int)` for extra amusement :) )

`bleach.clean(None)`, on the other hand, returns `u''` which is a more sensible of an answer.

However, for both of those functions, I would expect to either raise an exception or to return None, instead of returning an unicode string.

bleach==2.0.0
html5lib==0.999999999 (nine 9)

willkg:
I think to fix this, we'd want to change the beginning of `Cleaner.clean` and `Linker.link` to check if the input is a `six.text_type` and if not, raise a `TypeError` with an appropriate message.

I'm not going to work on this, but I'd accept a PR.

willkg:
@ivellios @overfl0 Either of you interested in working on this?

ivellios:
@willkg I will do. But... shouldn't it be rather `six.string_types` instead? I know it's time to move to Python 3.x, but with `text_type` it will not work for ASCII strings in Python 2.4+

ivellios:
Oh, and should I follow the release process: https://bleach.readthedocs.io/en/latest/dev.html#release-process ? Or just push changes with tests passing and let you handle the versioning, contributions and so on?

willkg:
@ivellios My bad--you're right. It should check for `six.string_types`.

Regarding the release process, that's only something I do--not you. If you want to fix this, just submit a PR covering the code change and tests and I can handle it from there.

Thank you!

willkg:
This is fixed now. Closing it out.

PR #301, opened 2017-08-26
Linkify and clean should accept only six.string_types

ivellios:
Linkify and clean should accept only six.string_types and otherwise now they raise TypeError.

This PR solves the issue: #292


willkg:
I've been swamped. I'll try to get to this soon. Possibly early September.

willkg:
This looks good. Thank you!

PR #306, opened 2017-09-19
Clarify security policy

### LINKED ISSUE #303: clarify security-issue status for Bleach
We have a proper security issue reporting process, but there's nothing in the documentation that covers what happens after that and what we promise.

This issue covers documenting that.

Plus we should probably go through CHANGES and make sure it points out security fixes if there were any. Pretty sure I did an audit a while back, but it's worth doing again.

willkg:
Fixes #303 

@g-k Do the changes to the README look ok to you?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,10 @@ Bleach Changes
 Version 2.0 (March 8th, 2017)
 -----------------------------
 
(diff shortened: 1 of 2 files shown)


willkg:
Thank you!

PR #312, opened 2017-09-20
Clean up tests

willkg:
This renames two of the test files to names that better reflect what they're
testing.

I also went through and adjusted the clean() tests to make it easier to read
them and verify the input and output. Granted, this is highly subjective and
specific to me, but I'm hoping the end result is that the tests are more uniform
in structure and easier to read for others, too.


willkg:
The github diff view of these changes is unhelpful. So I'll walk through what I did:

1. renamed `test_basics.py` to `test_clean.py` since that's what it's oding
2. nixed the `TestClean` class and dedented 90% of `test_clean.py` and removed `self` argument
3. adjusted tests where the text input to `clean` was hard to read, far away from the `bleach.clean()` call, or had a one-letter variable name such that the majority of the tests are structured similarly now and it's easier to see the input to `bleach.clean()` and the output together
4. renamed `test_link.py` to `test_linkify.py` since that's what it's doing

PR #319, opened 2017-09-20
prevent unescaping of entities (#143)

### LINKED ISSUE #143: HTML entities being unescaped

willkg:
This is a second attempt at fixing #143. It's based on the first attempt, but then adds a `sanitize_characters` method for sanitizing Characters tokens by extracting Entity tokens from the Characters token data. It's a bit tricky, but I think it's more correct than the previous attempt.

It doesn't escape all `&`, `clean` and `linkify` remain idempotent, and it's more backwards compatible.

Fixes #143.


willkg:
I want to run this through all the Standup output to see how it does. Plus I want to re-review the diff tomorrow.

willkg:
There are zero differences between Bleach 2.0 and this branch for Standup content. Standup uses both `bleach.clean` and `bleach.linkify` with some callbacks and other things with a Markdown parsing in the middle. It seems like character entities don't show up in the corpus often.

When they do, it's things like:

```
I did this & that today
```

which correctly gets cleaned to:

```
I did this &amp; that today
```

I'll re-review the diff tomorrow.

willkg:
This still looks ok to me, so I'm going to land it.

jvanasco:
thank you!

willkg:
@jvanasco Thank you! I really appreciate the work you did analyzing the previous attempt. That helped a lot.

PR #321, opened 2017-09-21
Fix character entities in HTML attributes

### LINKED ISSUE #294: escaping of URLs can causes issues in fragile context scenarios
_This issue is forked from discussion in #143 and #293_ 

Problems can arise from the plaintext escaping of URLs by the parser and how it relates to `linkify`. This is magnified when `linkify` is applied to `clean` text, but can also have issues when dealing with "replied" content. 
 
There are two variants of this issue, which concerns how the parser escapes the `&` character in a url into the HTML entity `&amp;` sequence of characters:

* general behaviors and "replied" content
* cases that involve a "fragile syntax" (using the W3C's terms)

General Behaviors

First, I apologize to @willkg for wasting your time on the `&amp;` encoding.  I was going by RFCs on HTTP and didn't dive into the HTML specs, which define that a browser should decode the `&amp;` and will request the "intended" url.  modern gui browsers all do this; commandline browsers seem hit-or-miss in catching this.  sending the html encoded url may or may-not work, as both `&` and `;` are often configured as the default delimiters (w3c and large projects seem to be standardizing on `&` only).  My bad.  

Fragile Syntax

This is the w3c example in which `&amp;` must be pre-escaped, since `&copy;` is a named character reference.  the link `?art&copy` will render as `art©`

    <a href="?art&amp;copy">Art and Copy</a> 

The existing code and current candidate are both deficient in handling this.

Here's a quick example script:

	html = '<a href="?art&amp;copy">Art and Copy</a>'
	print bleach.clean(html)
	print bleach.linkify(html)
	print bleach.linkify(bleach.clean(html))

	url = 'http://example.com?art&amp;copy'
	print bleach.clean(url)
	print bleach.linkify(url)
	print bleach.linkify(bleach.clean(url))

And the results on `v1.4.3` and the current #293

	+ ok
	x bad

	Bleach 1.4.3
	# html
	+ <a href="?art&amp;copy">Art and Copy</a>
	+ <a href="?art&amp;copy" rel="nofollow">Art and Copy</a>
	+ <a href="?art&amp;copy" rel="nofollow">Art and Copy</a>

	# url
	+ http://example.com?art&amp;copy
	x <a href="http://example.com?art©" rel="nofollow">http://example.com?art©</a>
	x <a href="http://example.com?art©" rel="nofollow">http://example.com?art©</a>

	# Pull 293
	# html
	x <a href="?art&amp;amp;copy">Art and Copy</a>
	+ <a href="?art&amp;copy" rel="nofollow">Art and Copy</a>
	x <a href="?art&amp;amp;copy" rel="nofollow">Art and Copy</a>

	# url
	+ http://example.com?art&amp;amp;copy
	+ <a href="http://example.com?art&amp;copy" rel="nofollow">http://example.com?art&amp;copy</a>
	x <a href="http://example.com?art&amp;amp;copy" rel="nofollow">http://example.com?art&amp;amp;copy</a>

willkg:
This is option 2 in 

https://github.com/mozilla/bleach/issues/294#issuecomment-331233053

I thought this would be more involved, but it wasn't. Go figure.

This is squirrely, but what's going on is that the tokenizer no longer consumes
entities so when the serializer goes to convert all & to &amp;, it's not a good
thing to do.

This wraps the serializer and looks for HTML attribute values, undoes the
categorical "all & shall be &amp;!" and then redoes it taking care only to
escape bare &.

Fixes #294.

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -153,9 +153,10 @@ class Cleaner(object):
 
         self.parser = BleachHTMLParser(namespaceHTMLElements=False)
         self.walker = html5lib.getTreeWalker('etree')
-        self.serializer = HTMLSerializer(
+        self.serializer = BleachHTMLSerializer(
             quote_attr_values='always',
             omit_optional_tags=False,
+            escape_lt_in_attrs=True,
 
             # We want to leave entities as they are without escaping or
             # resolving or expanding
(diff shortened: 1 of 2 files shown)


bleach/sanitizer.py
```diff
@@ -250,6 +251,79 @@ def _attr_filter(tag, attr, value):
...
+        "&"
+
+    """
+    for i, part in enumerate(AMP_SPLIT_RE.split(text)):
+        if i == 0:
+            yield part
+        elif i % 2 == 0:
+            yield '&' + part
+
+
```
willkg: All of this is the same as it was when it was methods in `BleachSanitizerFilter`. The only change here is that I pulled these two methods out of that class.

bleach/sanitizer.py
```diff
@@ -588,3 +591,57 @@ def sanitize_css(self, style):
...
+                    continue
+
+            yield part.replace('&', '&amp;')
+
+    def serialize(self, treewalker, encoding=None):
+        """Wrap HTMLSerializer.serialize and escape bare & in attributes"""
+        in_tag = False
+        after_equals = False
+
+        for stoken in super(BleachHTMLSerializer, self).serialize(treewalker, encoding):
```
willkg: `serialize` nicely returns strings one bit at a time. So:

```
<img src="?foo=bar">
```

Gets returned as the following parts which I'm calling "string tokens" or "stoken" (one per line):

```
<img
src
=
"
?foo=bar
"
>
```

We've got quoting set to "always", so an attribute value is always after a `=` followed by a `"`.

willkg:
Also, with this change, we're good again in regards to #294:

```
# string:  <a href="?art&amp;copy">Art and Copy</a>
+ clean:  <a href="?art&amp;copy">Art and Copy</a>
+ linkify:  <a href="?art&amp;copy" rel="nofollow">Art and Copy</a>
+ clean then linkify:  <a href="?art&amp;copy" rel="nofollow">Art and Copy</a>

# string:  http://example.com?art&amp;copy
+ clean:  http://example.com?art&amp;copy
+ linkify:  <a href="http://example.com?art&amp;copy" rel="nofollow">http://example.com?art&amp;copy</a>
+ clean then linkify:  <a href="http://example.com?art&amp;copy" rel="nofollow">http://example.com?art&amp;copy</a>
```

PR #323, opened 2017-09-28
Convert invisible characters to ? in Characters tokens

willkg:
This prevents someone from using backspace and other invisible characters from
tricking a user into copy and pasting a seemingly innocuous command into doing
something they really don't want to do.

I made the replacement character a constant figuring people can replace it if
they want something different.


willkg:
Once Travis is ok, I'm landing this and pushing out Bleach 2.1.

willkg:
We have a process for working on and reviewing security-related changes. This was reviewed there. Given that, I'm merging this.

Issue #341, opened 2017-12-15
Proposal: More control for style attribute.


sherubthakur:
Right now we can only pass in a list of css styles based on which we do the filtering. What opinion do you folks have on taking in a dict of styles instead and use it to filter out styles based on if they match or not.

i.e. right now we test cases like this would pass. (input, styles, expected)
```(
        'font-family: "Arial";',
        ['font-family'],
        'font-family: "Arial";'
    ),```

but if we can achieve something like 
```(
        'border: 1px solid blue; color: black;',
        {'font-family': '1px', 'color': 'black'},
        'color: black;'
    ),```

i.e. treating the values of the styles dict as regexs and only allowing styles that match the given regex. Support for functions would be also helpful.

Thoughts? If we do decide to do this I would be interested to work on it.

willkg:
I'm not sure I understand where you're going with this. From what you wrote, it sounds to me like you want to enforce style attribute property *and* values. Does that sound right?

What's the use case that you're trying to solve? Is it something like "I have difficulty expressing the list of style properties and values that we want to keep"? Or "we need to clean based on style property *and* value"? Or something else?

sherubthakur:
The use case I have in mind is I want to disallow/allow certain style values.

Why?
1. Certain styles can have values that can be used to do XSS but I do not want to outright filter them but have a regex or a function against those values which can check if these can values can lead to XSS and only filter those ones, and allow the rest.  
So from OWASP's XSS Filter Evasion Cheat Sheet background-image can be used to do XSS, but I would like to have control so that I can check `background-image` against a testing regex or a function and allow most values but block the ones that look invalid.
e.g. `<DIV STYLE="background-image: url(javascript:alert('XSS'))">`.

2. General filtering like I want to allow only certain values for some styles.

I my head it's like giving one more level of control to the user of this lib. @willkg thoughts?

willkg:
I can't tell what the compelling need is here. Is this a "nice-to-have" sort of thing or is it a "I need this functionality in the project I'm working on"? So far everything you've written feels like "nice-to-have".

Anyhow, thinking about the API here a bit, maybe it makes sense to follow `linkify` callbacks? Can we rework the existing CSS value sanitization as a set of functions and then let developers set up the ones they want and add their own? That feels easier to document and explain, and seems easier to deal with edge cases, than supporting a dict with values.

What do you think?

sherubthakur:
[...]

PR #342, opened 2017-12-15
Regex support for style value filtering

sherubthakur:
# WIP 
Regarding #341


willkg:
The conversation is continuing in #341. I think we should finish that up and if we decide to go with regexes, we can reopen this.

PR #343, opened 2017-12-18
Drop support for html5lib<1.0

hugovk:


willkg:
I appreciate you working on this!

I don't want to land this right now. I think in order to fix the regression issues, we're going to have to override large parts of the tokenizer. At that point, maybe it makes sense to vendor html5lib instead of depending on it. Also, I want to keep it super easy to do a CVE-fixit release and not have to deal with branches and mental models of what's where. Context-switching between big projects is hard.

Given all that, I don't want to land this now, so I'm going to let this sit for a bit. Maybe 3 months? It really depends on when I get some free time and/or what else comes up.

hugovk:
Replaced by #375.

willkg:
Oy--I said 3 months and it ended up being 6. Sorry about that. :(

PR #346, opened 2018-01-15
Add VALUE_SAFE return value for attribute filters

inducer:
In my use case (sanitize HTML output by student programs), I would like to allow certain `data:` URLs--specifically those of an `audio/wav` MIME type. Those currently (and rightfully) get killed by `allowed_protocols`. I'm wary of just outright allowing all `data:` URLs, and so I've extended the attribute filter interface with a special return value `VALUE_SAFE` that allows attribute values past without further vetting.

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -79,6 +79,13 @@ INVISIBLE_CHARACTERS_RE = re.compile(
 INVISIBLE_REPLACEMENT_CHAR = '?'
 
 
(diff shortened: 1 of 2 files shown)


willkg:
There's a lot to unpack here that's worth discussing. I suspect you can achieve what you're trying to achieve without making a substantial behavior change to Bleach.

Can you write up an issue in the issue tracker? We can discuss this further there.

Thank you!

PR #350, opened 2018-02-13
Support data URIs with content types whitelist. Based on html5lib.

iurisilvio:
https://github.com/html5lib/html5lib-python/blob/master/html5lib/filters/sanitizer.py#L820

To maintain backward compatibility, I didn't set any default content type. You must set `content_types` to use this feature.

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -11,6 +11,7 @@ from bleach.linkifier import (
 )
 from bleach.sanitizer import (
     ALLOWED_ATTRIBUTES,
+    ALLOWED_CONTENT_TYPES,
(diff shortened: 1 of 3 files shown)


willkg:
Hi! It's easier on me when people write up an issue with what they're trying to do and why and then write up a PR to make changes. That way we get to have a discussion and I understand where you're coming from.

Can you do that with this change, please?

iurisilvio:
Sure, I did it.

willkg:
Oops. I meant to close this out a while back. I wrote up this comment with why and a workaround:

https://github.com/mozilla/bleach/issues/351#issuecomment-416067178

PR #355, opened 2018-03-03
Fix MANIFEST and data_to_json.py related to recent changes

willkg:
I squashed test cases into single files--no more `.out` files. This carries
that change through to `MANIFEST.in` and our tests_website system.

I missed this in the previous cleanup passes. :(


willkg:
Thank you!

PR #361, opened 2018-03-19
Handle nonexistent namespaces better

### LINKED ISSUE #352: Exceptions on strange Html
Hello,

This small pattern raises an exception : 

    
    import bleach

    bleach.clean("<d {c}>")

willkg:
Issue 352 has a string that manages to tokenize an html attribute with
a namespace, but no name. Then the namespace doesn't exist in prefixes
and that throws a `KeyError`.

This alleviates that a bit such that if there's a namespace, but no
name, it swaps the two values. Further, if prefixes doesn't have the
namespace, then it ignores the namespace.

Fixes #352

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -668,8 +668,20 @@ class BleachSanitizerFilter(sanitizer.Filter):
             assert token_type in ("StartTag", "EmptyTag")
             attrs = []
             for (ns, name), v in token["data"].items():
+                # If we end up with a namespace, but no name, switch them so we
+                # have a valid name to use.
+                if ns and not name:
+                    ns, name = name, ns
+
+                # Figure out namespaced name if the namespace is appropriate
+                # and exists; if the ns isn't in prefixes, then drop it.
+                if ns is None or ns not in prefixes:
+                    namespaced_name = name
+                else:
+                    namespaced_name = '%s:%s' % (prefixes[ns], name)
+
                 attrs.append(' %s="%s"' % (
-                    name if ns is None else "%s:%s" % (prefixes[ns], name),
+                    namespaced_name,
                     # NOTE(willkg): HTMLSerializer escapes attribute values
                     # already, so if we do it here (like HTMLSerializer does),
                     # then we end up double-escaping.
--- a/tests/test_clean.py
+++ b/tests/test_clean.py
@@ -759,6 +759,19 @@ def test_convert_entities(data, expected):
     assert convert_entities(data) == expected
 
 
+def test_nonexistent_namespace():
+    """Verify if the namespace doesn't exist, it doesn't fail with a KeyError
+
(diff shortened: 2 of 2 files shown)


bleach/sanitizer.py
```diff
@@ -668,8 +668,20 @@ def disallowed_token(self, token):
             assert token_type in ("StartTag", "EmptyTag")
             attrs = []
             for (ns, name), v in token["data"].items():
+                # If we end up with a namespace, but no name, switch them so we
+                # have a valid name to use.
+                if ns and not name:
```
willkg: This is such a weird case. I don't see why the tokenizer should ever parse an html attribute as having a namespace, but no name. However, it's harder to fix this in html5lib than it is to just swap the values. So I swapped the values.

tests/test_clean.py
```diff
@@ -759,6 +759,19 @@ def test_convert_entities(data, expected):
...
+    """Verify if the namespace doesn't exist, it doesn't fail with a KeyError
+
+    The tokenizer creates "c" as a namespace and that doesn't exist in the map
+    of namespaces, so then it fails with a KeyError. I don't understand why the
+    tokenizer makes "c" into a namespace in this string.
+
+    Issue #352.
+
+    """
+    assert clean('<d {c}>') == '&lt;d c=""&gt;&lt;/d&gt;'
```
willkg: It's interesting to note that this expected outcome suffers from the regression between Bleach 1.5 and 2.0 where the tokenizer/parser add a `</d>` and then it gets escaped. Gross.

g-k: This is really weird to me since `d` isn't an html tag or in `ALLOWED_TAGS` and I'd expect the sanitizer to have dropped it at this point.

willkg: Yeah, so the tokenizer or the parser (I forget which) sees an html-like tag and then closes it off. Further, it removes `{` and `}` since those don't make sense in an attributes list and it adds a value to `c`. By the time it hits the Bleach sanitizer, it looks like:

```
<d c=""></d>
```

Then the Bleach sanitizer escapes the tag since it's not in the allow list turning it into that escaped nonsense.

If we passed `strip=True`, then it'd get stripped out completely:

```
$ python
Python 3.6.3 (default, Oct  3 2017, 21:45:48) 
[GCC 7.2.0] on linux
>>> import bleach
>>> bleach.clean('<d {c}>')
'&lt;d c=""&gt;&lt;/d&gt;'
>>> bleach.clean('<d {c}>', strip=True)
''
>>> 
```

That's what's going on. I agree it's weird. It'll get fixed when we fix the regression. I've been thinking about how to do that. I might tackle it in April along with some other things.

willkg:
The html5lib-python serializer needs to be fixed to support namespaces. Pretty sure it's covered in this issue: https://github.com/html5lib/html5lib-python/issues/59

I wrote up #362 to cover that.

willkg:
Thank you!

PR #364, opened 2018-03-22
Convert entities in CSS values before sanitizing

### LINKED ISSUE #363: "&quot;" in styles should be considered as a valid value
bleach 2.1.3

Style value like this `font-family: &quot;Droid Sans&quot;` is being treated as invalid.

However, this works in most browsers and is, in fact, a common practice.

willkg:
The CSS is in an HTML attribute value, so we need to convert character
entities in it which makes it proper CSS before we can sanitize it.

Fixes #363

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -702,7 +702,10 @@ class BleachSanitizerFilter(sanitizer.Filter):
 
     def sanitize_css(self, style):
         """Sanitizes css in style tags"""
-        # Drop any url values
+        # Convert entities in the style so that it can be parsed as CSS
+        style = convert_entities(style)
+
+        # Drop any url values before we do anything else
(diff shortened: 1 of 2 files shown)


willkg:
This converts character entities in a `style=` HTML attribute, so something like `&quot;` gets converted to `"`. We need to do that here because otherwise it's not valid CSS and so we can't parse it as CSS.

In other parts of the code, we try hard not to change it. Bleach is not an html parser, it's not a browser, it's just changing the text to make it HTML-safe. In this case, we are changing it by converting character entities. The serializer will escape `&`, `<`, and `>` in the HTML attribute value, so I think it's safe.

I think that's the best we can do.

Does that sound safe?

g-k:
Actually I need to think about this some more.

willkg:
Thank you!

PR #372, opened 2018-06-07
fixes 371 - reqs updates

### LINKED ISSUE #371: Upgrade Twine?
Hi! I'm a [Twine](https://twine.readthedocs.io) maintainer; thanks for using Twine. I noticed you've pinned an old version of Twine in [your requirements file](https://github.com/mozilla/bleach/blob/master/requirements.txt#L13). Twine [has changes in 1.11.0](https://twine.readthedocs.io/en/latest/changelog.html) that you might find useful, and I hope you'll upgrade. If you have any trouble upgrading, please do [let us know](https://github.com/pypa/twine/issues).

willkg:
This nixes pinning of development-centric packages for Bleach. There's no need for us to pin those packages (with the exception of pytest--we might want to pin that, but we can re-pin it later if we hit issues) and not pinning them reduces maintenance burden.

This also fixes the lint and docs tox environments which had bitrotted.

Fixes #371.


willkg:
Thank you!

PR #382, opened 2018-07-19
Add Python 3.7, htmllib 1.0b10 and 1.0.1 to tests

### LINKED ISSUE #377: add support for Python 3.7
Python 3.7 is out. We should make sure we support it.

jwhitlock:
Fixes #377, plus add the most recent htmlib versions. The new tox targets worked locally (thanks pyenv!)

Based on ``setup.py``, it looks like htmllib 0.99999999 (8) should be dropped from the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,13 @@ python:
 - "3.4"
 - "3.5"
 - "3.6"
+- "3.7"
 - "pypy"
 env:
 - HTML5LIB=0.99999999   # 8
 - HTML5LIB=0.999999999  # 9
+- HTML5LIB=1.0b10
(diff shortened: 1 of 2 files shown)


jwhitlock:
Oops, I misread the docs. Python 3.7 isn't available in TravisCI yet (https://github.com/travis-ci/travis-ci/issues/9815).

willkg:
Bleach supports 8 9s. It's encoded in setup.py in an odd way because of version comparison reasons.

Looks like Travis isn't testing the same matrix that we test with in tox. That's... annoying. I'll fix that now and add a note.

willkg:
I appreciate the effort! Thank you!

PR #383, opened 2018-07-20
Sync travis and tox environments

willkg:
This makes sure travis and tox are testing Bleach with the same
configurations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,21 @@
+# Note: If you update this, make sure to update tox.ini, too.
 sudo: false
 language: python
 cache:
   directories:
   - "~/.cache/pip"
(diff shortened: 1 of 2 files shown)


willkg:
@g-k When you get a chance, can you look this over?

willkg:
As a side note, there are some travis-tox integration libs out there, but this doesn't use one of them because I don't have any experience with them and didn't want to spend time looking into them right now. If someone else had experience with one of them and wanted to do a PR to set it up, that'd be super sweet.

willkg:
Thank you!

g-k:
Yeah, I don't see the configs changing frequently enough to warrant pulling in another library.

PR #387, opened 2018-08-26
vendor html5lib 1.0.1

### LINKED ISSUE #386: vendor html5lib
`pip` vendors html5lib and thus they don't have to support multiple html5lib APIs. Since html5lib 1.0.1 came out, there hasn't been much activity. We have issue #255 covering looking for alternatives. Any of those options is likely to be a big rewrite project.

In the meantime, we should vendor html5lib in the same way that pip does. That will simplify some code, reduce how long it takes to test, ease some problems with projects that use Bleach and other libraries that use html5lib, and generally make things better.

willkg:
This vendors html5lib 1.0.1 and in doing that, drops the requirement to
install html5lib.

Fixes #386


willkg:
This is kind of an insane thing to look at. I think I'd look at the following things:

1. `bleach/_vendor/README.rst` which lays out the policy for vendoring libraries. Is the language clear? Are there additional policy points we should make/enforce?
2. `bleach/_vendor/pip_install_vendor.sh` which creates the vendored packages. This is probably fine.

I think if those two things look ok, then we're ok. The rest is effectively generated code.

bleach/__init__.py
```diff
@@ -18,20 +17,6 @@
...
-    _html5lib_version = html5lib.__version__.split('.')
-    if len(_html5lib_version) < 2:
-        _html5lib_version = _html5lib_version + ['0']
-except Exception:
-    _h5ml5lib_version = ['unknown', 'unknown']
-
-
-# Bleach 3.0.0 won't support html5lib-python < 1.0.0.
-if _html5lib_version < ['1', '0'] or 'b' in _html5lib_version[1]:
-    warnings.warn('Support for html5lib-python < 1.0.0 is deprecated.', DeprecationWarning)
```
g-k: Removing this import madness is satisfying.

bleach/_vendor/vendor.txt
```diff
@@ -0,0 +1 @@
+html5lib==1.0.1
```
g-k: I think we should have hashes here.

willkg:
^^^ That adds a hash for html5lib (good idea!) and adds some review instructions.

I think we can add another test to Travis/Tox that verifies the integrity of the vendored libraries. I just have to think about how that might work. Probably install the libraries into a `/tmp` directory and then diff the two directory trees?

I'll tinker with that and add it to this PR.

willkg:
I tweaked travis and tox and now both run vendorverify and lint more correctly. I verified that Travis is running vendorverify now.

@g-k How's that last commit look?

.travis.yml
```diff
@@ -4,18 +4,30 @@ language: python
...
   - pip install -r requirements-dev.txt
+
+matrix:
+  include:
+    - python: 2.7
+      env: MODE=lint
+    - python: 2.7
+      env: MODE=vendorverify
+    - python: 3.4
+      env: MODE=lint
```
willkg: Turns out we weren't running lint in Travis. This fixes that as well as running vendorverify.

We lint in both 2.7 and 3.4 because flake8 does different things between Python 2 and 3.

.travis.yml
```diff
@@ -4,18 +4,30 @@ language: python
...
+
 script:
-  - py.test
-  - flake8 bleach/
+  - ./scripts/run_tests.sh $MODE
```
willkg: This switches to more of a "dispatch" style. The end result is that we're not running vendorverify and lint for every possibility.

scripts/run_tests.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Make sure we're running from the bleach repository directory and
+# not this directory.
+THISDIR=$(basename `pwd`)
+if [[ "${THISDIR}" == "scripts" ]]; then
+    cd ..
+fi
```
willkg: Tox runs commands via subprocess, so it can't run a command in a different directory without switching to that directory first (or us adding a bunch of shenanigans). So then we need this script to run both from the scripts directory as well as the bleach repository root. This bit figures out what the working directory is and fixes it if it's the wrong one.

willkg:
Thank you so much! This will make things soooo much easier going forward.

g-k:
yeah thanks for doing this!

PR #388, opened 2018-08-27
Drop easy_install instructions

### LINKED ISSUE #373: drop easy_install instructions
`easy_install` is done and we don't test the installation instructions we have for that. We should drop any `easy_install` related instructions.

willkg:
This isn't earth-shattering or anything. I haven't tested `easy_install` installation ever, so I don't even know if it works anymore.

Fixes #373.


willkg:
Thank you!

PR #389, opened 2018-08-28
Extract all html5lib things into a shim module

### LINKED ISSUE #316: move html5lib override code into separate "html5lib shim" module
We're increasingly subclassing html5lib classes and overriding behavior and that code is getting tangled with Bleach's behavior code. I'm concerned this is getting worse over time and it'll become increasingly more difficult to upgrade it to support new versions of html5lib.

I think it behooves us to centralize and move the html5lib shim code into a separate module with a "defined API" and then change the Bleach code to sit on top of that.

willkg:
This creates an intermediary `html5lib_shim` module between the html5lib
library and Bleach. This will make it easier to update html5lib and fix
any issues that arise from that.

We're going to treat this as an incompatible API change, so changing
the upcoming version to 3.0.0.

Fixes #316

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -1,0 +1,258 @@
+# flake8: noqa
+"""
+Shim module between Bleach and html5lib. This makes it easier to upgrade the
+html5lib library without having to change a lot of code.
+"""
+
+from __future__ import unicode_literals
+
+import re
+import string
+
+import six
+
+from bleach._vendor.html5lib import (
+    HTMLParser,
+    getTreeWalker,
+)
+from bleach._vendor.html5lib.constants import (
+    entities,
+    namespaces,
+    prefixes,
+    tokenTypes,
+)
+from bleach._vendor.html5lib.constants import _ReparseException as ReparseException
+from bleach._vendor.html5lib.filters.base import Filter
+from bleach._vendor.html5lib.filters.sanitizer import allowed_protocols
+from bleach._vendor.html5lib.filters.sanitizer import Filter as SanitizerFilter
+from bleach._vendor.html5lib.serializer import HTMLSerializer
+from bleach._vendor.html5lib._tokenizer import HTMLTokenizer
+from bleach._vendor.html5lib._trie import Trie
+
+
+#: Map of entity name to expanded entity
+ENTITIES = entities
+
+#: Trie of html entity string -> character representation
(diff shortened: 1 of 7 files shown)


tests/test_html5lib_shim.py
```diff
@@ -0,0 +1,23 @@
...
+    ('&#x20;', ' '),
+
+    # Handles ambiguous ampersand
+    ('&xx;', '&xx;'),
+
+    # Handles multiple entities in the same string
+    ('this &amp; that &amp; that', 'this & that & that'),
+])
+def test_convert_entities(data, expected):
+    assert html5lib_shim.convert_entities(data) == expected
```
willkg: Possible fixme: more of the functions should have tests in this module. They're implicitly tested via other tests in `test_clean`, but maybe it's prudent to test them explicitly here?

willkg:
Relatedly, this is all to pave the way to make it easier to fix the big regression which I'm pretty sure we have to fix in the tokenizer. It's still going to be a HUGE pain in the ass because of how the tokenizer is constructed, but at least now it'll be "easier".

tests/test_html5lib_shim.py
g-k: If `convert_entities` isn't a public API, I'd say keep it DRY and don't duplicate the tests as long as coverage stays the same. Otherwise, might be worth duplicating the tests.

bleach/html5lib_shim.py
```diff
@@ -0,0 +1,258 @@
...
+        stream.pop(0)
+
+        if stream and stream[0] in ('x', 'X'):
+            allowed = '0123456789abcdefABCDEF'
+            possible_entity += stream.pop(0)
+        else:
+            allowed = '0123456789'
+
+        # FIXME(willkg): Do we want to make sure these are valid number
+        # entities? This doesn't do that currently.
```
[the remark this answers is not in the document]
willkg: This was a FIXME I added a while ago when I originally implemented this. I don't see anything in html5lib, but I'm not sure I know what form it would have. I have no idea what risk it would have offhand.

tests/test_html5lib_shim.py
willkg: That makes sense. I'll think about it a bit more and see if I can think of a compelling reason one way or the other.

willkg:
^^^ That improves some comments and adds some tests for the serializer. After that passes, I'll land it.

willkg:
Thank you!

PR #390, opened 2018-09-04
Fix ids for regression tests

willkg:
This fixes test ids so when the regression tests fail, it's easier to isolate and run just one of them. Otherwise, this is cosmetic.

### DIFF
--- a/tests/test_clean.py
+++ b/tests/test_clean.py
@@ -757,10 +757,10 @@ def test_nonexistent_namespace():
     assert clean('<d {c}>') == '&lt;d c=""&gt;&lt;/d&gt;'
 
 
-def get_tests():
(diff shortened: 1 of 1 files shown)


tests/test_clean.py
```diff
@@ -772,14 +772,20 @@ def get_tests():
...
+        for fn in tests
     ]
 
     return testcases
 
 
-@pytest.mark.parametrize('fn, test_case', get_tests())
-def test_regressions(fn, test_case):
+_regression_ids_and_tests = get_ids_and_tests()
+_regression_ids = [item[0] for item in _regression_ids_and_tests]
```
willkg: regression ids ends up being a list of just the file names. For example:

```
['1.test', '2.test', ...]
```

PR #391, opened 2018-09-13
Fix regressions from Bleach 2.0 rewrite

### LINKED ISSUE #271: unclosed script tag is escaped incorrectly
`bleach.clean('<ul><li><script></li></ul>')` used to (correctly, I would assume,) just escape the `<script>` and output `<ul><li>&lt;script&gt;</li></ul>`

now the output is `<ul><li>&lt;script&gt;&lt;/li&gt;&lt;/ul&gt;&lt;/script&gt;</li></ul>`

Previously we were using Bleach 1.5.0 with html5lib 0.9999999; now we're using Bleach 2.0.0 with[ html5lib 0.999999999](https://pyup.io/changelogs/html5lib/#0.999999999)

### LINKED ISSUE #279: <isindex> tag being replaced with a string
This seems to be a new thing in 2.0.0. In 1.5.0 the below snippet would return a blank string.

```python
>>> from bleach import clean
>>> clean('<isindex>', tags=[], strip=True)
'This is a searchable index. Enter search keywords: '
```

Is this intentional? Any way to tell bleach to strip it completely? I realize ``<isindex>`` is a deprecated tag, but it's present in data we're trying to clean coming from old third party software.

### LINKED ISSUE #280: tags get added when tokenizing with bleach 2.0
When I redid Bleach to support html5lib >= 0.99999999 (8 9s), I rewrote the sanitizing code to operate as an html5lib filter rather than as a tokenizer. The result of this is that the tokenizer cleans up the HTML before Bleach gets a chance to work on it and the act of cleaning up the HTML sometimes changes or creates new tags.

I wrote up this issue: https://github.com/html5lib/html5lib-python/issues/312

For example, if you had:
```
import bleach
bleach.clean('Yeah, sure. <sarcasm/>', strip=False)
```
it'd get turned into:
```
'Yeah, sure. &lt;sarcasm&gt;&lt;/sarcasm&gt;.'
```
and then that'd get escaped/dropped/whatever.

Adding this as an issue to help catch other regressions and use cases so I can figure out if I want to spend more time looking into a fix.

willkg:
The Bleach 2.0 rewrite involved changing Bleach `clean()` from making
changes during the tokenizing step to operating as an html5lib filter
that runs between the parsing and serializing steps. There were some
consequences of that rewrite namely that the html5lib parser got to
"fix" the HTML before Bleach could strip/escape disallowed tags. Because
of that, there were a variety of regressions from Bleach 1.x behavior
and Bleach 2.x behavior. For example:

1. Added end tags:
   ```
   bleach.clean('<sarcasm>') -> '&lt;sarcasm&gt;&lt;/sarcasm&gt;'
   ```
2. Parsed malformed tags as comments:
   ```
   bleach.clean('</sarcasm>') -> ''
   ```
Et cetera.

This change fixes that in a few ways. First, it overrides `emitCurrentToken()`
in the tokenizer to know about the allowed tags and whether they're going
to get stripped or escaped. It does the stripping at this point and changes
tags to be escaped into character data. That way the parser doesn't do any
"fixing".

Second, it adds some additional scrutiny for the
"expected-closing-tag-but-got-char" parser error case. That gets kicked
up when the tag is malformed. The parser will treat this as a comment
and then Bleach `clean()` will drop it. Instead, this fixes that to
turn that into character data to be escaped later.

Fixes #271
Fixes #279
Fixes #280


bleach/html5lib_shim.py
```diff
@@ -37,8 +38,156 @@
...
+        self._buffer.extend(list(chars))
+        return chars
+
+    def unget(self, char):
+        if self._buffer:
+            self._buffer.pop(-1)
+        return self._inner_stream.unget(char)
+
+    def stream_history(self):
+        return self._buffer
```
willkg: This literally just wraps input streams and keeps track of what they've emitted so far. This has the adverse effect of keeping the characters in the stream in memory, but Bleach tends to work on strings that are smallish, so I think this is "ok" for "common use cases".

It's possible we could figure out a way to keep track of just the interesting part of the stream history. Like everything from the most recent `<` that isn't in quotes. That requires some additional finagling and complexity, so I decided not to do that for now.

bleach/html5lib_shim.py
```diff
@@ -37,8 +38,156 @@
...
+        # Stomp on the HTMLTokenizer __init__ in order to wrap the stream.
+        self.stream = InputStreamWithMemory(HTMLInputStream(stream, **kwargs))
+
+        # Do all the things the HTMLTokenizer does in __init__
+        self.parser = parser
+        self.escapeFlag = False
+        self.lastFourChars = []
+        self.state = self.dataState
+        self.escape = False
+        self.currentToken = None
```
willkg: This `__init__` is the same as `HTMLTokenizer` except it wraps the input stream with one that remembers the history of the stream.

tests/test_clean.py
```diff
@@ -710,20 +740,6 @@ def test_svg_allow_local_href_nonlocal(text, expected):
...
-
-@pytest.mark.xfail(reason='regression from bleach 1.5')
-def test_sarcasm():
-    """Jokes should crash.<sarcasm/>"""
-    assert (
-        clean('Yeah right <sarcasm/>') ==
-        'Yeah right &lt;sarcasm/&gt;'
-    )
-
-
```
willkg: I merged a bunch of tests into a big `stripping` test with many test cases and a big `escaping` test with many test cases. It cleaned up this file a bunch plus added some additional coverage.

willkg:
I ran this through the Standups status message corpus (58,000+ messages some of which are from people working on stylo and other HTML tag things) and it looks **much** better now. I didn't hit anything new that should get fixed.

bleach/html5lib_shim.py
```diff
@@ -51,9 +200,55 @@ def consumeEntity(self, allowedChar=None, fromAttribute=False):
...
+                # If we're stripping the token, we just throw in an empty
+                # string token.
+                new_data = ''
+
+            else:
+                # If we're escaping the token, we want to escape the exact
+                # original string. Since tokenizing also normalizes data
+                # and this is a tag-like thing, we've lost some information.
+                # So we go back through the stream to get the original
+                # string and use that.
```
g-k: comments :+1:

bleach/html5lib_shim.py
```diff
@@ -51,9 +200,55 @@ def consumeEntity(self, allowedChar=None, fromAttribute=False):
...
+                'type': tokenTypes['Characters'],
+                'data': new_data
+            }
+
+            self.currentToken = new_token
+            self.tokenQueue.append(new_token)
+            self.state = self.dataState
+            return
+
+        super(BleachHTMLTokenizer, self).emitCurrentToken()
```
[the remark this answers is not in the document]
g-k: Nice!

I'd worry about escaping somehow leading to: 1) dangling tags (an opening tag without a closing one and potentially leaking data), 2) extra tags, or 3) escaped characters that get ignored but result in an HTML tag.

However, I don't think that should be possible if if `[StartTag] to `[Characters]` is escaping things properly for an HTML document/body context. I think the caution is justified though. I'll have to think about this some more.

bleach/html5lib_shim.py
```diff
@@ -37,8 +38,156 @@
...
+        self.escapeFlag = False
+        self.lastFourChars = []
+        self.state = self.dataState
+        self.escape = False
+        self.currentToken = None
+
+    def __iter__(self):
+        last_error_token = None
+
+        for token in super(BleachHTMLTokenizer, self).__iter__():
```
g-k: nit: maybe makes sense to cache the `supert(BleachHTMLTokenizer, self)` since we're using it in a few places? However, it's nice to have the clarity of seeing where's it's used directly too.

bleach/html5lib_shim.py
```diff
@@ -51,9 +200,55 @@ def consumeEntity(self, allowedChar=None, fromAttribute=False):
...
+            return
+
+        super(BleachHTMLTokenizer, self).emitCurrentToken()
+
 
 class BleachHTMLParser(HTMLParser):
     """Parser that uses BleachHTMLTokenizer"""
+    def __init__(self, tags, strip, **kwargs):
+        """
+        :arg tags: list of allowed tages--everything else is either stripped or
```
g-k: nit: typo `tages`

bleach/html5lib_shim.py
```diff
@@ -51,9 +200,55 @@ def consumeEntity(self, allowedChar=None, fromAttribute=False):
...
+            escaped; if None, then this doesn't look at tags at all
+        :arg strip: whether to strip disallowed tags (True) or escape them (False);
+            if tags=None, then this doesn't have any effect
```
g-k: Would it make sense to have separate parsers for strip and escape modes?

tests/data/10.test
```diff
@@ -1,3 +1,3 @@
 <IMG SRC="javascript:alert('XSS');">
 --
-&lt;img src="javascript:alert('XSS');"&gt;
+&lt;IMG SRC="javascript:alert('XSS');"&gt;
```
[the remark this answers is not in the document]
g-k: Oh, nice

tests/data/3.test
```diff
@@ -1,3 +1,3 @@
 >"'><img%20src%3D%26%23x6a;%26%23x61;%26%23x76;%26%23x61;%26%23x73;%26%23x63;%26%23x72;%26%23x69;%26%23x70;%26%23x74;%26%23x3a;alert(%26quot;%26%23x20;XSS%26%23x20;Test%26%23x20;Successful%26quot;)>
 --
-&gt;"'&gt;&lt;img%20src%3d%26%23x6a;%26%23x61;%26%23x76;%26%23x61;%26%23x73;%26%23x63;%26%23x72;%26%23x69;%26%23x70;%26%23x74;%26%23x3a;alert(%26quot;%26%23x20;xss%26%23x20;test%26%23x20;successful%26quot;)&gt;&lt;/img%20src%3d%26%23x6a;%26%23x61;%26%23x76;%26%23x61;%26%23x73;%26%23x63;%26%23x72;%26%23x69;%26%23x70;%26%23x74;%26%23x3a;alert(%26quot;%26%23x20;xss%26%23x20;test%26%23x20;successful%26quot;)&gt;
+&gt;"'&gt;&lt;img%20src%3D%26%23x6a;%26%23x61;%26%23x76;%26%23x61;%26%23x73;%26%23x63;%26%23x72;%26%23x69;%26%23x70;%26%23x74;%26%23x3a;alert(%26quot;%26%23x20;XSS%26%23x20;Test%26%23x20;Successful%26quot;)&gt;
```
g-k: :smile:

tests/test_clean.py
```diff
@@ -140,32 +140,14 @@ def test_comments(data, should_strip, expected):
...
-    # Test with braces
-    ('a <style>body{}</style> test', 'a &lt;style&gt;body{}&lt;/style&gt; test'),
-])
-def test_disallowed_tags(data, expected):
-    assert clean(data) == expected
-
-
 def test_invalid_char_in_tag():
-    # NOTE(willkg): Two possible outcomes because attrs aren't ordered
```
g-k: yesss

tests/test_clean.py
```diff
@@ -745,16 +761,11 @@ def test_invisible_characters(data, expected):
...
-
-    Issue #352.
-
-    """
-    assert clean('<d {c}>') == '&lt;d c=""&gt;&lt;/d&gt;'
+    # Issue #352 involved this string kicking up a KeyError since the "c"
+    # namespace didn't exist. After the fixes for Bleach 3.0, this no longer
+    # goes through the HTML parser as a tag, so it doesn't tickle the bad
+    # namespace code.
+    assert clean('<d {c}>') == '&lt;d {c}&gt;'
```
g-k: :+1: :100:

willkg:
@g-k These comments are awesome! I'll work through them over the next couple of days.

bleach/html5lib_shim.py
willkg: A maxsize might be interesting, but if it fails you end up with missing text and it's unclear why. I think I want to punt on that.

I don't think a deque helps here. The usage of `self._buffer` is in `get_recent_tag_string` where it gets reversed and then turned into a list then the list is iterated over. Using a deque doesn't really help except possibly in the `unget` case but I don't think that triggers often. It's possible that we could rewrite `get_recent_tag_string` in a faster/more-performant fashion. I didn't spend much time thinking about that. If we rewrote it, then a deque might help more.

bleach/html5lib_shim.py
willkg: Oh, that's a really good idea. I think I didn't do that because there was an intermediary period where this was doing more. I forget.

I'll fix this now.

bleach/html5lib_shim.py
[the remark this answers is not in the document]
willkg: Just to clarify, this covers `StartTag` and `EndTag`, so we should never end up with dangling tags. In both cases, the tokenizer will check to see if they're allowed and if not, convert them to `Characters`.

bleach/html5lib_shim.py
```diff
@@ -37,8 +38,156 @@
...
+    def stream_history(self):
+        return self._buffer
+
+
+def get_recent_tag_string(stream_history, token):
+    """Find the original text for the tag
+
+    This goes back through the stream we've tokenized for the most recent
+    complete HTML tag-like thing as it existed in the stream. It assumes that
+    the current character in the stream ias a >.
```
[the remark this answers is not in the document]
willkg: Oops!

bleach/html5lib_shim.py
willkg: That's really interesting. It would make sense if it reduced the likelihood of mistakes and/or was more performant. I'll write up an issue to look into that after I land this.

bleach/html5lib_shim.py
willkg: It occurs to me that once `get_recent_tag_string` has been called we know that the stream is a bunch of stuff up to the most recent tag. We'll never want to return anything in that stream history again, so we can dump it. That'll reduce the memory issue.

bleach/html5lib_shim.py
willkg: `super` walks the MRO for the thing to call and it should be pretty fast. The MRO won't change, so we could cache it, but I'm not sure a name lookup would be much faster than the work to look it up in the MRO. Maybe it is? Maybe it's something that's changed over Python versions? I'm pretty sure dicts got a speed boost recently.

I think for now I'm going to leave it as is.

bleach/html5lib_shim.py
willkg: Oops!

willkg:
^^^ That fixes/addresses many of the comments.

Work that still needs to get done:

* [ ] look at effect of non-well-formed HTML in `get_recent_tag_string` (https://github.com/mozilla/bleach/pull/391#discussion_r218428516)
* [ ] think about reimplementing `get_recent_tag_string` using a regex (https://github.com/mozilla/bleach/pull/391#discussion_r218430195)
* [ ] write up issue to look into separate escape/strip tokenizers

willkg:
^^^ That nixes `get_recent_tag_string()` altogether by abusing `tagOpenState()`. This is a muuuuuch better idea.

willkg:
I think this is as good as we're going to get it at this stage. I'm going to land it and solicit help testing it out.

Thank you, @g-k!

PR #394, opened 2018-10-02
fixes #392: linkify fix

### LINKED ISSUE #392: linkify() "fixes" HTML by adding or removing tags
Similar to the issue in https://github.com/mozilla/bleach/issues/280, `bleach.linkify()` seems to be "fixing" HTML by adding or removing tags if it interprets something as an HTML tag.

For example:

```
>>> bleach.linkify("insert <name> here")
'insert <name> here</name>'

>>> bleach.linkify("Well that's just great</sarcasm>")
"Well that's just great"
```

willkg:
This tweaks the code for the `BleachHTMLParser` and `BleachHTMLTokenizer` to work with the needs of the linkifier. This fixes the case where the linkifier was "fixing" non-HTML tags like `</sarcasm>`.

Fixes #392.

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -38,8 +38,9 @@ ENTITIES = entities
 ENTITIES_TRIE = Trie(ENTITIES)
 
 #: Token type constants--these never change
-START_TAG_TYPE = tokenTypes['StartTag']
-END_TAG_TYPE = tokenTypes['EndTag']
+TAG_TOKEN_TYPES = set([
+    tokenTypes['StartTag'], tokenTypes['EndTag'], tokenTypes['EmptyTag']
+])
 CHARACTERS_TYPE = tokenTypes['Characters']
 
 
@@ -158,10 +159,10 @@ class BleachHTMLTokenizer(HTMLTokenizer):
             self.tokenQueue.append({"type": CHARACTERS_TYPE, "data": '&'})
 
     def tagOpenState(self):
-        # This state marks a < that is either a StartTag, EndTag, or ParseError.
-        # In all cases, we want to drop any stream history we've collected
-        # so far and we do that by calling start_tag() on the input stream
-        # wrapper.
+        # This state marks a < that is either a StartTag, EndTag, EmptyTag,
+        # or ParseError. In all cases, we want to drop any stream history
+        # we've collected so far and we do that by calling start_tag() on
+        # the input stream wrapper.
         self.stream.start_tag()
         return super(BleachHTMLTokenizer, self).tagOpenState()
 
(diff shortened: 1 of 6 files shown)


bleach/html5lib_shim.py
```diff
@@ -32,15 +31,31 @@
...
+HTML_TAGS = [
+    tag for namespace, tag in
+    (
+        list(constants.scopingElements) +
+        list(constants.formattingElements) +
+        list(constants.specialElements) +
+        list(constants.htmlIntegrationPointElements) +
+        list(constants.mathmlTextIntegrationPointElements)
+    )
+]
```
willkg: This builds a list of all known HTML tags that html5lib knows about.

I thought about making this an argument, but that requires some new naming and tests and documenting work. I currently think if a user needs to specify tags, then they should be sanitizing and then linkifying.

bleach/html5lib_shim.py
```diff
@@ -139,17 +156,23 @@ def __iter__(self):
...
 
             yield token
 
     def consumeEntity(self, allowedChar=None, fromAttribute=False):
-        # We don't want to consume and convert entities, so this overrides the
-        # html5lib tokenizer's consumeEntity so that it's now a no-op.
+        # If this tokenizer is set to consume entities, then we can let the
+        # superclass do its thing.
+        if self.consume_entities:
+            return super(BleachHTMLTokenizer, self).consumeEntity(allowedChar, fromAttribute)
```
willkg: When sanitizing, the sanitizer is doing the entity work. The linkifier isn't doing any sanitization, so we want it to do whatever the tokenizer originally did.

bleach/linkifier.py
```diff
@@ -110,9 +110,16 @@ def __init__(self, callbacks=DEFAULT_CALLBACKS, skip_tags=None, parse_email=Fals
...
+        # anything not in that list.
+        self.parser = html5lib_shim.BleachHTMLParser(
+            tags=html5lib_shim.HTML_TAGS,
+            strip=False,
+            consume_entities=True,
+            namespaceHTMLElements=False,
+        )
         self.walker = html5lib_shim.getTreeWalker('etree')
-        self.serializer = html5lib_shim.HTMLSerializer(
+        self.serializer = html5lib_shim.BleachHTMLSerializer(
```
willkg: This is the big change. We switch the Linkifier to use our tokenizer and parser. That accounts for known tags vs. unknown tags and does the right thing in regards to escaping. Because we're using our tokenizer and parser, we need to use our serializer, too, otherwise we get multiple rounds of escaping and linkify isn't idempotent anymore.

docs/linkify.rst
```diff
@@ -5,12 +5,20 @@
...
+HTML links.
+
+It works by parsing the text as HTML and building a document tree. In this
+way, it's guaranteed never to do weird things to URLs in attribute values,
+can modify the value of attributes on ``<a>`` tags and can even do things
+like skip ``<pre>`` sections.
+
+If you plan to sanitize/clean the text and linkify it, you should do that
+in a single pass using :ref:`LinkifyFilter <linkify-LinkifyFilter>`. This
+is faster and it'll use the list of allowed tags from clean.
```
willkg: I tried to improve this documentation so it's clearer that if you're going to linkify, you probably want to do it when sanitizing.

willkg:
Awesome--thank you!

PR #401, opened 2018-10-05
Add abbr to HTML_TAGS

### LINKED ISSUE #400: abbr is escaped when using linkify
From IRC:

```
willkg: I'm doing the work to upgrade AMO to bleach3.0 but I may have found a
regression with the linkify fix (?).
`bleach.linkify('<abbr title="laugh out loud">lol</abbr>', skip_tags=['abbr'])` escapes
the `<abbr>` tag.  Is there a way to have linkify _not_ do that?
```

Looks like `abbr` isn't in `HTML_TAGS`. We should add it.

willkg:
This fixes the issue where linkify will escape `<abbr>`.

Fixes #400

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -56,6 +56,8 @@ HTML_TAGS = [
         list(constants.mathmlTextIntegrationPointElements)
     )
 ]
+# Add tags that aren't in html5lib.constants
+HTML_TAGS.extend(['abbr'])
 
 
 class InputStreamWithMemory(object):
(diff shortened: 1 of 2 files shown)


bleach/html5lib_shim.py
```diff
@@ -56,6 +56,8 @@
         list(constants.mathmlTextIntegrationPointElements)
     )
 ]
+# Add tags that aren't in html5lib.constants
+HTML_TAGS.extend(['abbr'])
```
willkg: I did the `extend` here figuring we're going to probably add others in the future and might as well make it easier.

willkg:
Thank you!

PR #404, opened 2018-10-07
Fix regression with dynamic Cleaner object tags.

### LINKED ISSUE #403: Changing Cleaner.tags dynamically stopped working in 3.0.0
Previously in version 2.1.4 I would be able to do something like this:

```
import bleach
cleaner = bleach.Cleaner()
cleaner.tags.append("script")
```

Now in version 3.0.0, it seems like there was a refactor which causes `self.tags` to be used during `__init__` (to build html5lib parcer) so the appended tag doesn't propagate to other composed objects.

russellballestrini:
Fixes #403

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -123,12 +123,6 @@ class Cleaner(object):
         self.strip_comments = strip_comments
         self.filters = filters or []
 
-        self.parser = html5lib_shim.BleachHTMLParser(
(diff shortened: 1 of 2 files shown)


willkg:
Per https://github.com/mozilla/bleach/issues/403#issuecomment-427856139, I don't want to support this right now, so I'm going to close it out.

Thank you for the contribution! Sorry it didn't work out!

PR #405, opened 2018-10-08
fix 377 - add python 3.7 support

willkg:
This fixes `vendor_verify.sh` script accounting for `RECORD` being unsorted and occasionally failing the diff.

This also adds testing Bleach in Python 3.7 to tox and Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,15 @@ install:
 
 matrix:
   include:
-    - python: 2.7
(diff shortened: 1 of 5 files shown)


willkg:
Bleach tests passed in Python 3.7 in Travis CI build so that's cool. I think we're good here.

willkg:
Thank you!

Issue #409, opened 2018-10-09
linkify is escaping valid HTML tags


Deimos:
Originally noticed here in relation to the `<abbr>` tag: https://github.com/mozilla/bleach/issues/400

linkify is now using a list of valid HTML tags that is largely built up from constants in html5lib, here: https://github.com/mozilla/bleach/blob/master/bleach/html5lib_shim.py#L49

However, these constants don't seem to make up a complete list of valid tags&mdash;in addition to `<abbr>` (which is now being appended to the list by Bleach), there are also multiple other tags missing, including `del`, `ins`, `sub`, `sup`, and `span`.

It would probably be best to replace this combination of html5lib constants with a full list of valid tags. I can do a PR for this.

willkg:
For anyone who's blocked on this, you can go back to the original pre-3.0 linkify behavior by using `LinkifyFilter` in your own parsing pipeline.

It's something like this:

```
from bleach import html5lib_shim
from bleach.linkify import Linkifyfilter


class OldLinker:
        def __init__(self, callbacks=DEFAULT_CALLBACKS, skip_tags=None, parse_email=False,
            url_re=URL_RE, email_re=EMAIL_RE):
        self.callbacks = callbacks
        self.skip_tags = skip_tags
        self.parse_email = parse_email
        self.url_re = url_re 
        self.email_re = email_re 
        self.parser = html5lib_shim.HTMLParser(namespaceHTMLElements=False) 
        self.walker = html5lib_shim.getTreeWalker('etree') 
        self.serializer = html5lib_shim.HTMLSerializer(
            quote_attr_values='always', 
            omit_optional_tags=False, 
            # linkify does not sanitize
            sanitize=False,
            # linkify alphabetizes
            alphabetical_attributes=False,
        )

    def linkify(self, text):
        dom = self.parser.parseFragment(text)
        filtered = LinkifyFilter(
            source=self.walker(dom),
            callbacks=self.callbacks,
            skip_tags=self.skip_tags,
            parse_email=self.parse_email,
            url_re=self.url_re,
            email_re=self.email_re,
        )
        return self.serializer.render(filtered)
```

willkg:
With #412 merged, I want to additionally tweak the `Linker` to have an additional argument specifying the list of tags it knows about. I'll do that in a followup PR.

PR #412, opened 2018-10-11
Define a full list of valid elements for HTML_TAGS

Deimos:
Previously, the list of valid HTML tags being used by linkify was a
combination of various element lists from the html5lib constants.
However, these lists don't cover all of the valid HTML elements, so
linkify was escaping some valid tags (including `<abbr>`, `<span>`, and
more).

This commit just defines a full list of valid, non-deprecated HTML
elements for linkify to use instead.

This should address https://github.com/mozilla/bleach/issues/409, please let me know if you'd like this to be done in a different manner.


bleach/html5lib_shim.py
```diff
@@ -45,19 +45,119 @@
 CHARACTERS_TYPE = constants.tokenTypes['Characters']
 
 
-#: List of HTML tags
+#: List of valid HTML tags
```
willkg: If we're going to have a comprehensive list of tags, we should specify how it's comprehensive. Is this the complete list of tags defined in the HTML 5.3 spec or something else?

Deimos: Sure - I had taken it from MDN originally, but I've updated it slightly now to use the list from the WHATWG spec, and added a link to their list in their index. (I'm not sure what the `#:` comments indicate, so I don't know if my new line should have a colon prefix too).

Let me know how that seems or if you'd like something else.

ismikeup:
Linkify has had a showstopping issue (and people interested in addressing it) for almost two months now. Has there movement on this recently? When can we expect a release that addresses #412?

willkg:
@ismikeup It's in my list of things to do, but I haven't gotten there, yet. Sorry!

ismikeup:
Is there anything that I can do to help?

willkg:
Part of the issue that I want to work out before I land this is what happens when user's want a different set of tags. In the html5lib project, there are people complaining about issues with mathml and svg tags. Some templating languages are "html-like" and maybe those should get accounted for, too. I don't have a good feel for who's doing what and what the use cases we should cover should be.

Anyhow, I'd rather linkify didn't grow an additional "all the tags linkify should know about" argument which would conflict with using linkify by itself vs. clean with linkify, but inevitably, someone is going to want a different list of tags--maybe they're doing mathml or svg or some templating language. I've been tossing around not landing this, but instead doing a different approach that's more flexible and doesn't involve me maintaining a massive allow list of HTML tags.

We hang out on #bleach on irc.mozilla.org if you or anyone else want to show up and chat about it.

ismikeup:
Bleach is an incredibly important project, I'm glad you're thoughtful. 

My issue with `linkify` has always been that the documentation makes using it seem very straightforward. And it is.. Until you hit a snag! I'm 100% in favor of exposing more of the underlying complexity to users. 

In my use case I'm cleaning the HTML and sticking it in a cache in one step and later I'm pulling it out, linkifying it based on current context, and presenting it to a user. Being able to white list tags and attrs in `linkify` the way we're able to with `clean` would be enormously useful.

willkg:
You've got a very specific easy-to-solve use case because you already have an allow list. You can write your own Linker which is based on this:

https://github.com/mozilla/bleach/blob/bdb28650a2ffe7a994f59b4a958e8b8763b53b64/bleach/linkifier.py#L73

Adjust that to pass in the allow list in the `__init__` and pass that into the `BleachHTMLParser` and you should be all set.

ismikeup:
That's what I ended up doing a few weeks ago. Feels like it's a common case that could be supported out of the box.

willkg:
What common case are you talking about? Is it "clean and cache the result, then linkify much later"? Or is it "linkify with a specified allowed set of tags"?

Also, given that it's been a month and a half and only like 5 people have interacted with the issue and pull request, I'm thinking it's not common at all. That's an important thing because it does affect how I prioritize the myriad of things on my plate. So if there's other people who are just sitting around waiting for this to happen but haven't chimed in in any way, I have no idea.

ismikeup:
Yes, linkify with control over which tags are allowed. I believe that the ability to linkify text without escaping certain tags is important. Being able to do that without writing a linker would be convenient. 

Convince aside, the fact that some tags are escaped by default is not disclosed in the `linkify` documentation. That's what took me, and seems to have taken others, by surprise.

willkg:
I think we're off topic at this point. For any outstanding concerns, please open up a new issue.

I will get to this at some point. I work on Bleach in spurts because that's what my time allows. If that doesn't work for you, consider participating in development. Sorry I can't do better!

ismikeup:
That tags (and which specific ones) are being escaped is very much on topic. 

No apology necessary, your commitment (time or otherwise) is not being questioned or taken for granted. Your work is high quality and massively appreciated. No need to take a defensive posture. 

As stated, I would love to get involved (https://github.com/mozilla/bleach/pull/412#issuecomment-441734705). Will take your recommendation and hop on IRC.

Issue #234, opened 2016-11-03
Support cleaning HTML for presentational purposes


EmilStenstrom:
In #67 there was a discussion about adding more attributes to the clean method to allow stripping the contents of disallowed tags, and not only the contents. I think this would be a great idea, and would allow us to remove other cleaning we do today.

Our use-case is that we receive unsafe e-mail messages. Some of them are full HTML documents, but most are just HTML fragments. They are extremely messy, and we want to pick out the content and a few well-specified formatting tags from the contents of that HTML. Using clean() doesn't work since it keeps the contents of for instance `<script>` tags, so we've implemented a separate cleaning step that uses lxml.html.clean. Cleaner for this. This is not specific to script tags, style tags are also affected.

In #67 it was suggested that this new methods was called `prettify`. What's needed to continue the discussion if this is a good idea?

dangayle:
Also here: https://github.com/mozilla/bleach/issues/185

dangayle:
This is important. Stripping unsafe tags is a safety feature, I get that, but it's gross to leave the javascript text (or other unwanted text node) still visible. Without this, we would either have to roll our own or use an entirely different library to accomplish both tasks, which undermines the usefulness of bleach.

tgs:
The filter API in html5lib seems pretty useful here.  `<style>` and `<script>` are special because they can't have child tags, so the logic can be very simple:

```python
import sys
import html5lib
from html5lib.filters.base import Filter


class NoChildTagFilter(Filter):
    """
    ONLY WORKS ON TAGS THAT CANNOT HAVE CHILDREN
    """
    def __init__(self, stream, tags):
        Filter.__init__(self, stream)
        self._tags = [t.lower() for t in tags]

    def __iter__(self):
        tokens = Filter.__iter__(self)
        while True:
            for token in tokens:
                if token['type'] == 'StartTag' and token['name'].lower() in self._tags:
                    break
                yield token
            else:
                # we ran out of tokens
                break
            for token in tokens:
                if token['type'] == 'EndTag' and token['name'].lower() in self._tags:
                    break


if __name__ == '__main__':
    dom = html5lib.parseFragment(sys.stdin.read(), treebuilder="dom")
    walker = html5lib.getTreeWalker("dom")
    stream = walker(dom)

    s = html5lib.serializer.HTMLSerializer()

    sys.stdout.write(s.render(NoChildTagFilter(stream, ('script', 'style'))))
```

[...]

PR #413, opened 2018-10-16
initial attempt at stripping content of script and style tags

jvanasco:
this is an attempt at addressing #234 (removing the contents of script and style tags)

a new option is added to several methods: `tags_strip_content`. this is a list of tags which should have their content stripped.

the stripping is achieved in two phases:

1. The tokenizer will ignore these tags.
2. The serializer maintains state of being in a tag marked for stripping

I had tried to do an alternate implementation in which:

* the tokenizer would escape/strip the tag as in the current implementation
* the tokenizer adds a payload to the token noting that it should be stripped

unfortunately that approach is incompatible with multiple parts of the tag parsing. there may be a way to convert the tags to something else (perhaps `<unsafe>` then strip out the unsafe tags on render.

this could be refactored into a new 'prettify' function as previously suggested, but it seems cleaner as a kwarg

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -29,7 +29,7 @@ __all__ = ['clean', 'linkify']
 
 def clean(text, tags=ALLOWED_TAGS, attributes=ALLOWED_ATTRIBUTES,
           styles=ALLOWED_STYLES, protocols=ALLOWED_PROTOCOLS, strip=False,
-          strip_comments=True):
+          strip_comments=True, tags_strip_content=None, ):
     """Clean an HTML fragment of malicious content and return it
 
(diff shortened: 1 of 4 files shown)


willkg:
I think you could implement this as a filter. The tags that you want to remove the contents of can be in the list of allowed tags for clean and then the filter removes them and all their contents.

Have you tried something like that?

jvanasco:
> The tags that you want to remove the contents of can be in the list of allowed tags for clean 

FFS, why did I not see that?

Yep.  That works.

Are you open to a PR that adds `filters` to `bleach.clean`?

willkg:
I'm open to adding additional filters that people can use if they so choose--like `LinkifyFilter`.

PR #414, opened 2018-10-16
filtering tags and their contents 

jvanasco:
this could be broken up into two PRs/concepts.

1. the main `bleach.clean` function is extended with a filters argument, which simply proxies the value to `Cleaner.__init__`

2. an `extras.py` was added to support a `TagTreeFilter`, which is a filter capable of removing an entire tag tree.  as a reference, utility and test-requirement a function named `clean_strip_content` was also included to invoke this.  This function addresses #234

For the concern of security/safety, this filter deletes the entire tree of a matching tag - including any tags that may appear within it.

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -29,7 +29,7 @@ __all__ = ['clean', 'linkify']
 
 def clean(text, tags=ALLOWED_TAGS, attributes=ALLOWED_ATTRIBUTES,
           styles=ALLOWED_STYLES, protocols=ALLOWED_PROTOCOLS, strip=False,
-          strip_comments=True):
+          strip_comments=True, filters=None):
     """Clean an HTML fragment of malicious content and return it
 
     This function is a security-focused function whose sole purpose is to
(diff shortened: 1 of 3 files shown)


willkg:
I'm still a "weak no" to adding prettification-related things to Bleach.

There have been a handful of issues and I understand the use cases, but Bleach is an HTML sanitization security library and prettification isn't in scope. I'm not convinced that prettification features in Bleach are generally valuable to the Bleach user base. I don't see enough value yet to offset the additional maintenance burden and potential increase in attack surface.

I highly encourage you and others to start a new library that meets your prettification needs. Maybe it contains a bunch of Bleach filters. Maybe it adds some additional glue. Maybe it adds some documentation for common things to do. Figure out what use cases it should solve and what it shouldn't solve. Figure out how it should work and how users should use it. Prove its worth. At that point, we can talk and maybe merge it in. Maybe start with these changes?

jvanasco:
i disagree with stripping the contents not being security focused, but it's not worth arguing over.  I originally wrote that as a standalone for internal use, so just released that github and pypi - https://github.com/jvanasco/bleach_extras

HOWEVER

how do you feel about the first item I brought up though -- extending `bleach.clean` with a `filters` argument that is just passed on to Cleaner.  those are the only edits to `__init__.py` on the changeset above.

I do think that change should happen, as it's the only difference between `clean()` and `Cleaner.__init__`.  i'll gladly generate a new PR with only that change.

willkg:
`bleach.clean` is a shortcut and I don't really want to keep adding more to it. I'd prefer to remove it since it's far slower than creating your own `Cleaner`.

So for now, I'm not interested in adding that argument to `bleach.clean`.

PR #423, opened 2018-12-12
Add recognized_tags to Linker arguments

### LINKED ISSUE #409: linkify is escaping valid HTML tags

willkg:
`recognized_tags` lets someone specify a different set of recognized tags
to the `Linker` when they're using it by itself. One use of this is if
you're doing a clean pass, storing that, and then linkifying on demand.
The `recognized_tags` argument lets you re-use the allowed tags from the
clean step and then there isn't any additional oddities occuring in
the linkify step.

Fixes #409

### DIFF
--- a/bleach/linkifier.py
+++ b/bleach/linkifier.py
@@ -85,7 +85,7 @@ class Linker(object):
 
     """
     def __init__(self, callbacks=DEFAULT_CALLBACKS, skip_tags=None, parse_email=False,
-                 url_re=URL_RE, email_re=EMAIL_RE):
+                 url_re=URL_RE, email_re=EMAIL_RE, recognized_tags=html5lib_shim.HTML_TAGS):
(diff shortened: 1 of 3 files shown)


willkg:
Thank you!

PR #429, opened 2019-01-03
Convert &amp; to & as a Characters token

### LINKED ISSUE #422: LinkifyFilter with clean breaks links at &amp;
(Split out to a new issue from [this comment](https://github.com/mozilla/bleach/issues/419#issuecomment-445056894))

When using `LinkifyFilter` instead of the stand-alone `linkify`, links are broken at any `&amp;`, but not bare `&` characters.

Here's a pretty minimal demonstration:

```
import bleach

bleach.linkify("https://www.youtube.com/watch?v=SkcucKDrbOI&amp;feature=youtu.be")
```

Result (expected): `<a href="https://www.youtube.com/watch?v=SkcucKDrbOI&amp;feature=youtu.be" rel="nofollow">https://www.youtube.com/watch?v=SkcucKDrbOI&amp;feature=youtu.be</a>`

```
import bleach
from functools import partial

bleach_linkifier = partial(bleach.linkifier.LinkifyFilter)
cleaner = bleach.Cleaner(filters=[bleach_linkifier])
cleaner.clean("https://www.youtube.com/watch?v=SkcucKDrbOI&amp;feature=youtu.be")
```

Result (not expected, link stops at the `&amp;`): `<a href="https://www.youtube.com/watch?v=SkcucKDrbOI">https://www.youtube.com/watch?v=SkcucKDrbOI</a>&amp;feature=<a href="http://youtu.be">youtu.be</a>`

It works with just a bare `&` instead of the `&amp;`, like `https://www.youtube.com/watch?v=SkcucKDrbOI&feature=youtu.be`. In that case, the output of both methods is:

```
<a href="https://www.youtube.com/watch?v=SkcucKDrbOI&amp;feature=youtu.be">https://www.youtube.com/watch?v=SkcucKDrbOI&amp;feature=youtu.be</a>
```

(with an additional `rel="nofollow"` attr added by `linkify()` by default)

willkg:
This fixes a problem in `LinkifyFilter` when using it with the `Cleaner` where
the `Cleaner` sets up the tokenizer to not consume entities. So character
entities end up in their own `Entity` tokens and `Linkifyfilter` can't match
links that cross token boundaries. If there's a `&amp;`, then `LinkifyFilter`
won't match across that.

This fixes that by converting `&amp;` to `&` in the sanitizer when it's pulling out
entities and putting them in separate `Entity` tokens. The `&` `Characters` tokens
will get merged by `BleachSanitizerFilter.__iter__` and `&` will get converted
back to `&amp;` in the serialier.

Fixes #422

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -395,7 +395,18 @@ class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
             if part.startswith('&'):
                 entity = html5lib_shim.match_entity(part)
                 if entity is not None:
-                    new_tokens.append({'type': 'Entity', 'name': entity})
+                    if entity == 'amp':
+                        # LinkifyFilter can't match urls across token boundaries
+                        # which is problematic with &amp; since that shows up in
+                        # querystrings all the time. This special-cases &amp;
+                        # and converts it to a & and sticks it in as a
+                        # Characters token. It'll get merged with surrounding
+                        # tokens in the BleachSanitizerfilter.__iter__ and
+                        # escaped in the serializer.
+                        new_tokens.append({'type': 'Characters', 'data': '&'})
+                    else:
+                        new_tokens.append({'type': 'Entity', 'name': entity})
+
                     # Length of the entity plus 2--one for & at the beginning
                     # and and one for ; at the end
                     remainder = part[len(entity) + 2:]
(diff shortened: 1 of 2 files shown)


willkg:
These fiddly problems drive me crazy. This one took hours to work out. I think this is the best approach. I looked at a few others involving modifying the tokenizer, the parser, and wrapping `LinkifyFilter` in a `merge_charaters` kind of iterator. All those options had various problems from being hard to reason about or adding another performance hit or I couldn't make them work. Ugh.

One nice thing about this approach is that if we hit other character entities that need to be expanded, we can do that in this section and it'll "just work".

If you see anything better, I'm all ears.

willkg:
Thank you!

PR #433, opened 2019-01-08
Fix doctest failures

### LINKED ISSUE #357: "make doctest" kicks up lots of failures
Running "make doctest" in either Python 2 or Python 3 yields a bunch of failures.

This issue covers figuring out why and fixing it and then updating the release process to specify which version of Python to run them in.

willkg:
This reworks the doctests to run and pass in Python 3.

Fixes #357

### DIFF
--- a/docs/clean.rst
+++ b/docs/clean.rst
@@ -63,10 +63,10 @@ For example:
    >>> import bleach
 
    >>> bleach.clean(
-   ...     u'<b><i>an example</i></b>',
+   ...     '<b><i>an example</i></b>',
    ...     tags=['b'],
    ... )
-   u'<b>&lt;i&gt;an example&lt;/i&gt;</b>'
+   '<b>&lt;i&gt;an example&lt;/i&gt;</b>'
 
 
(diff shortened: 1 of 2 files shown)


docs/clean.rst
```diff
@@ -353,7 +353,7 @@ Trivial Filter example:
 .. doctest::
 
    >>> from bleach.sanitizer import Cleaner
-   >>> from html5lib.filters.base import Filter
+   >>> from bleach.html5lib_shim import Filter
```
willkg: This one is interesting. What's going on is that we vendored `html5lib`, so it's no longer installed as a requirement. That's cool, but that means it's not installed anymore. I switched it to use the `bleach.html5lib_shim` module. It's the same `Filter` class, but it's specifically the one that ships with Bleach. That's more correct. I'm not sure if it feels weird to be importing it from a module named `bleach.html5lib_shim` or not. It's probably fine.

g-k: wfm

willkg:
Thank you!

Issue #248, opened 2017-02-21
redo sanitize_css


willkg:
html5lib has this issue for replacing parts of the `sanitize_css()` method with a better CSS parser:

https://github.com/html5lib/html5lib-python/issues/152

Bleach has its own `sanitize_css()` which probably has some of the same problems, but not all of them. Might be worth switching to a library instead.

willkg:
I'm putting this in the 2.0 milestone for now because I'm changing lots of things, but will push it out if I don't have time.

willkg:
I got tinycss sort of working with Bleach. Something like this:

```
    def sanitize_css(self, style):
        """Sanitizes css in style tags"""
        parsed, errors = self.css_parser.parse_style_attr(style)

        if not parsed:
            return u''

        def decl_to_css(decl):
            pri = '!' + decl.priority if decl.priority else ''
            return '%s: %s%s' % (decl.name, decl.value.as_css(), pri)

        parsed = [
            decl
            for decl in parsed
            if (decl.name.lower() in self.allowed_css_properties or
                decl.name.lower() in self.allowed_svg_properties)
        ]

        if not parsed:
            return u''

        return u'; '.join([decl_to_css(decl) for decl in parsed]) + u';'
```

We have a test that verifies that bogus CSS gets dropped:

```
def test_valid_css():
    """The sanitizer should fix missing CSS values."""
    styles = ['color', 'float']
    assert (
        clean('<p style="float: left; color: ">foo</p>', styles=styles) ==
        '<p style="float: left;">foo</p>'
    )
    assert (
        clean('<p style="color: float: left;">foo</p>', styles=styles) ==
        '<p style="">foo</p>'
    )
```

It fails the second assertion in that test. For some reason tinycss thinks `color: float: left;` is valid and lets it through. I'm not sure whether that's important or not.

I think I'm going to leave things as they are and think about this issue some more another time. Possible options:

1. fix that bug upstream
2. look at tinycss2: https://github.com/SimonSapin/tinycss2
3. find another library
4. write our own parser that does the bare minimum for what we need to do--we kind of have one already

willkg:
Oo--supporting CSS property functions is pretty interesting. Maybe `sanitize_css` should get adjusted to work like attributes do where it can take a callable for filtering?

g-k:
@peterbe suggested some other options too:

* https://github.com/ebook-utils/css-parser
* https://bitbucket.org/cthedot/cssutils/src/default/

[...]

PR #434, opened 2019-01-09
fix #426

### LINKED ISSUE #426: Style filter does not work correctly when supplied non-ASCII style attribute
# Environment

Windows 10 Pro
Python 3.7.1
bleach         3.0.2

I've setup so that `color` style will survive after clean() method like below:
```
>>> bleach.clean('<span style="font-family:consolas;color:red">text</span>', tags = ['span'], attributes = {'span': ['style']}, styles = ['color'])
'<span style="color: red;">text</span>'
```

# Problem

When it contains non-ASCII style description such as Japanese font names, the `color` style is omitted.
```
>>> bleach.clean('<span style="font-family:メイリオ;color:red">text</span>', tags = ['span'], attributes = {'span': ['style']}, styles = ['color'])
'<span style="">text</span>'
```
It should retain the `color` style like below:
```
'<span style="color: red;">text</span>'
```

sekineh:
This PR will fix #426.

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -593,7 +593,7 @@ class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
         # the whole thing.
         parts = style.split(';')
         gauntlet = re.compile(
-            r"""^([-/:,#%.'"\sa-zA-Z0-9!]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$"""
+            r"""^([-/:,#%.'"\sa-zA-Z0-9!\w]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$"""
         )
 
         for part in parts:
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -593,7 +593,7 @@ class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
         # the whole thing.
         parts = style.split(';')
         gauntlet = re.compile(
-            r"""^([-/:,#%.'"\sa-zA-Z0-9!\w]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$""",
+            r"""^([-/:,#%.'"\s!\w]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$""",
             flags=re.U
         )
 
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -593,7 +593,8 @@ class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
         # the whole thing.
         parts = style.split(';')
         gauntlet = re.compile(
-            r"""^([-/:,#%.'"\sa-zA-Z0-9!\w]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$"""
+            r"""^([-/:,#%.'"\sa-zA-Z0-9!\w]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$""",
+            flags=re.U
         )
 
         for part in parts:
--- a/tests/test_css.py
+++ b/tests/test_css.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from functools import partial
 
 import pytest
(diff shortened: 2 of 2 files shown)


bleach/sanitizer.py
```diff
@@ -593,7 +593,7 @@ def sanitize_css(self, style):
         # the whole thing.
         parts = style.split(';')
         gauntlet = re.compile(
-            r"""^([-/:,#%.'"\sa-zA-Z0-9!]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$"""
+            r"""^([-/:,#%.'"\sa-zA-Z0-9!\w]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$"""
```
sekineh: I could instead eliminate the whole following lines because it has no corresponding tests:

https://github.com/mozilla/bleach/blob/8d7fd48179b5020d9b1521be7b81e06648d868d3/bleach/sanitizer.py#L590-L601

At this first attempt, I took more conservative path.

sekineh:
Hmm, I can't write utf-8 literal in test py files.

sekineh:
In python2, `\w` seems to have different meaning than python3.
It seems regex is not a reliable method for sanitizing something.

sekineh:
@willkg I think it's done.
Could you please review this?

tests/test_css.py
```diff
@@ -70,11 +76,19 @@ def test_allowed_css(data, styles, expected):
...
+            p_double = unicode(p_double)
         assert clean(p_double.format(data), styles=styles) == p_double.format(expected)
     else:
+        if python2_and_unicode(data):
+            p_single = unicode(p_single)
         assert clean(p_single.format(data), styles=styles) == p_single.format(expected)
 
 
+def python2_and_unicode(data):
+    return not isinstance(data, str)
```
willkg: The function name suggests this should return `True` if it's running in Python 2 and this is unicode, but the test seems only to test if the data is `str`. I'm not sure what this is trying to do, but it seems like a mismatch.

Maybe something like this is better?:

```
def is_python2_str(data):
    return six.PY2 and isinstance(data, str)
```

bleach/sanitizer.py
```diff
@@ -593,7 +593,8 @@ def sanitize_css(self, style):
         # the whole thing.
         parts = style.split(';')
         gauntlet = re.compile(
-            r"""^([-/:,#%.'"\sa-zA-Z0-9!]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$"""
+            r"""^([-/:,#%.'"\sa-zA-Z0-9!\w]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$""",
```
willkg: For some backstory, this gauntlet regex is mediocre and I'm not even sure I understand what it's really doing. Is it enforcing valid declaration structure? Doesn't seem like it is. It originally came from html5lib and I think it's been tweaked periodically since then, but I'd bet it doesn't correctly match CSS declarations. Issue #248 covers replacing this with tinycss or tinycss2. I started looking into that, but then hit some issues and stopped.

Looks like you added `\w` which will match word characters and `_` in that first part. I think that's fine, but I wonder why it wasn't like that in the first place. Regardless, I think you can remove the `a-zA-Z0-9` and `-` in the first part. So then the regex is like this:

```
             r"""^([/:,#%.'"\s!\w]|\w-\w|'[\s\w]+'\s*|"[\s\w]+"|\([\d,%\.\s]+\))*$""", 
```

tests/test_css.py
```diff
@@ -14,6 +15,11 @@
         ['color'],
         'color: red;'
     ),
+    (
+        u'font-family: メイリオ; color: red; float: left; background-color: red;',
```
willkg: So far, we've been keeping these files as ASCII and doing unicode code points for non-ASCII characters. I think we should do that here, too. We can either do that in this PR or I can do it afterwards--either is fine with me.

bleach/sanitizer.py
sekineh: done in 33f294d

tests/test_css.py
sekineh: done in 33f294d.
Used the name `is_python2_unicode` which matches to the required logic.

sekineh: btw, I didn't know the `six` library. Thank you for suggestion!

tests/test_css.py
sekineh: I've left this.

sekineh:
@willkg I added a commit to satisfy your request!

willkg:
Thank you!

PR #437, opened 2019-01-18
Fix style tests

willkg:
This fixes the style tests so they're more explicit and less
composed-on-the-fly which makes them harder to verify.


willkg:
Thank you!

PR #438, opened 2019-02-13
Unvendor the html5lib library

jdufresne:
html5lib development has stabilized in recent months. So concerns of frequent API breakage is less than it was in the past.

Reduces the need for users to have html5lib installed multiple times: once by itself and once by Bleach. setup.py now uses a version specifier to ensure a compatible version is installed.

Unvendoring is more secure. Should the html5lib library discover and fix a security issue, library users can update html5lib -- and only html5lib -- to incorporate the fix across their project. In Bleach's current
state, Bleach will immediately need to make a release to include such a security fix. This creates unnecessary additional attack vectors.


willkg:
I made the decision to vendor html5lib. I also spearheaded the html5lib 1.0 release. While I appreciate your arguments, I still think vendoring html5lib was the right thing to do for Bleach. html5lib has no security bug reporting process. The maintainers are unpaid, unappreciated, and tired. Development has stalled. Given that, html5lib is effectively an unmaintained project. Vendoring it makes it a lot easier for us to fix security issues that come up. Thus, I claim Bleach is in a better position to handle security issues quickly and effectively--not a worse one. Vendoring Bleach happened in issue #386.

I stepped down as Bleach maintainer, so this is no longer my decision to make, but I don't think anything substantive has changed in the situation to make me reconsider vendoring it at this time.

g-k:
I enabled dependabot for the `bleach/_vendor` directory (which picks up `vendor.txt` as a requirements file), so we'll get security updates for html5lib (possibly through github vuln notifications too).

I'm interested in attack vectors that vendoring html5lib introduces in addition to a delayed time to patch providing a longer exploitation window.

Happy to reconsider if html5lib gets the funding and security bug reporting process @willkg mentioned, but I'm closing this for now.

claudep:
Looks like html5lib maintainership is in better shape now. I cannot say much about security reporting, but is this vendoring issue worth being reconsidered again?

PR #443, opened 2019-04-12
Remove u string prefix in files with unicode_literals

jdufresne:
Using `from __future__ import unicode_literals` guarantees that all string literals will be Unicode strings by default. Using the u prefix is redundant. Cleans up the code towards a Python 3 style.

### DIFF
--- a/bleach/callbacks.py
+++ b/bleach/callbacks.py
@@ -3,31 +3,31 @@ from __future__ import unicode_literals
 
 
 def nofollow(attrs, new=False):
-    href_key = (None, u'href')
+    href_key = (None, 'href')
 
     if href_key not in attrs:
         return attrs
 
-    if attrs[href_key].startswith(u'mailto:'):
+    if attrs[href_key].startswith('mailto:'):
         return attrs
 
-    rel_key = (None, u'rel')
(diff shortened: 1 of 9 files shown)


jdufresne:
> We should also add the futures import and drop u' from test files (from git grep u\' | grep tests/), if you want to do that in this PR that'd be cool otherwise happy to open an issue for it.

No problem. Done in the latest revision.

g-k:
Sweet thanks @jdufresne!

Also if you're interested, we're looking for maintainers.

jdufresne:
> Also if you're interested, we're looking for maintainers.

Sure, I can help out here and there. Probably not full time though if that is alright.

I recently started integrating Bleach with one of my projects, which is why I began looking at the code.

g-k:
> Sure, I can help out here and there. Probably not full time though if that is alright.

@jdufresne that'd be great. The time commitment is definitely not full time and up to you. I spend maybe a hour or two a month on bleach, so I'll probably ping you for additional feedback on the occasional issue or PR. Bleach is stable and willkg wrapped up the major changes before stepping down, so it should be pretty quiet. Thanks!

PR #444, opened 2019-04-12
Simplify tox configuration

jdufresne:
Remove boilerplate comment from the top of the file.

In envlist, move pypy to same line as rest of the default Python
environments.

Don't override basepython. The default uses the correct Python
executable:

https://tox.readthedocs.io/en/latest/config.html#conf-basepython

> If not specified, the virtual environments factors (e.g. name part)
> will be used to automatically set one. For example, py37 means
> python3.7, py3 means python3 and py means python.


g-k:
Thanks!

PR #448, opened 2019-04-13
Fix ResourceWarning emitted during tests

jdufresne:
Always close the open file by using a context manager. Previously
Appeared as:

    tests/test_clean.py:790
      tests/test_clean.py:790: ResourceWarning: unclosed file <_io.TextIOWrapper name='.../tests/data/1.test' mode='r' encoding='UTF-8'>
        for fn in tests

### DIFF
--- a/tests/test_clean.py
+++ b/tests/test_clean.py
@@ -785,10 +785,11 @@ def get_ids_and_tests():
     # Sort numerically which makes it easier to iterate through them
     tests.sort(key=lambda x: int(os.path.basename(x).split('.', 1)[0]))
 
-    testcases = [
-        (os.path.basename(fn), open(fn, 'r').read())
-        for fn in tests
-    ]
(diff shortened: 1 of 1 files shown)


g-k:
Thank you!

tests/test_clean.py
```diff
@@ -785,10 +785,11 @@ def get_ids_and_tests():
     # Sort numerically which makes it easier to iterate through them
     tests.sort(key=lambda x: int(os.path.basename(x).split('.', 1)[0]))
 
-    testcases = [
-        (os.path.basename(fn), open(fn, 'r').read())
-        for fn in tests
-    ]
+    testcases = []
+    for fn in tests:
+        with open(fn) as fp:
```
g-k: note to self: open defaults to read mode / 'r' so this open call is equivalent

Issue #450, opened 2019-04-22
bleach.clean method returns unexpected value for string with unknown entities


aliaksandr-trush:
For string strings like "&adp;" and "&*!;" actual output is "&ad;;" and "&*;;".
But I expect that such values should remain original value ("&adp;" and "&*!;" corresponding).

g-k:
Hey @aliaksandr-trush! Thanks for reporting this. `bleach.clean` does not retain the original values for entities html5lib doesn't recognize (see https://github.com/mozilla/bleach/blob/77e1acdb863e59a1b6494ffa6e9d304a79f0403c/bleach/sanitizer.py#L395 and https://github.com/mozilla/bleach/blob/77e1acdb863e59a1b6494ffa6e9d304a79f0403c/bleach/_vendor/html5lib/constants.py#L651 for the list of recognized entities).

Are there specific values you expect it to retain? Generally speaking we'd like clean should remove or escape unparseable HTML.

alvinlindstam:
I took a look at this, and found a possible bug, I attempted a fix at https://github.com/mozilla/bleach/pull/452

I noticed that the library is not using the `resolve_entities` and `consume_entities` features of html5lib, and implements it's own html entity sanitizer. Do you know why that is? What difference is it that we want to achieve compared to letting them parse the entities and produce valid entity output?

willkg:
> Do you know why that is? What difference is it that we want to achieve compared to letting them parse the entities and produce valid entity output?

html5lib is an html5 parser and *consumes* the entities it recognizes. When we rewrote Bleach to use the Filter API, that caused a bunch of regressions from prior Bleach behavior. So I wrote a bunch of code to override those parts of the html5lib parser and tokenizer so they didn't consume entities and let Bleach sanitize properly.

willkg:
I think the problem is here:

https://github.com/mozilla/bleach/blob/47fe22e72dd4b2f54b1dd6bccfdc3eee212675c5/bleach/html5lib_shim.py#L499-L506

That consumes from the stream until it the thing it's assembling is definitely not a known character entity and the character it's going to look at is not an end character.

With `&adp;`, we have this:

```
   stream               c     possible_entity
0  ['d', 'p', ';']      a 
1  ['p', ';']           d     a
2  [';']                p     ad
```

[...]

PR #452, opened 2019-05-01
Issue 450: invalid character entities

alvinlindstam:
Possible fix for https://github.com/mozilla/bleach/issues/450.

This change fixes the case where there was precisely two characters between a valid character entity prefix and a semicolon, and the helper function returned the entity `ad` from `&adp;` (which later led to returning `&ad;;`.

In doing so, I also changed so that other invalid character entities (such as `&xx;`) were escaped, instead of keeping them around if the non-matchning parts were short. Not sure what you think about that?

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -488,7 +488,7 @@ def match_entity(stream):
         while stream and stream[0] not in end_characters:
             c = stream.pop(0)
             if c not in allowed:
-                break
+                return None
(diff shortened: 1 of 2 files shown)


willkg:
This hasn't gone anywhere in a while. Per https://github.com/mozilla/bleach/issues/450#issuecomment-488385656 , I'm going to close it out. If someone wants to tackle fixing it, we can do it in a new PR.

PR #458, opened 2019-07-05
Check parser.tags for None on "expected-closing-tag-but-got-char"

bope:
Hi,

I hit this edge case while using `BleachHTMLParser` with `tags=None`. `BleachHTMLTokenizer` will try to iterate `self.parser.tags` if `last_error_token['data'] == 'expected-closing-tag-but-got-char'`.

I've added a test along with the fix. I also had to update the exception string checks in `test_clean.py` since they've changed with pytest 5 update (11 hours ago?).

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -256,7 +256,7 @@ class BleachHTMLTokenizer(HTMLTokenizer):
                     yield token
 
                 elif ((last_error_token['data'] == 'expected-closing-tag-but-got-char' and
-                     token['data'].lower().strip() not in self.parser.tags)):
+                     (self.parser.tags is None or token['data'].lower().strip() not in self.parser.tags))):
                     # We've got either a malformed tag or a pseudo-tag or
                     # something that html5lib wants to turn into a malformed
                     # comment which Bleach clean() will drop so we interfere
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -256,7 +256,8 @@ class BleachHTMLTokenizer(HTMLTokenizer):
                     yield token
 
                 elif ((last_error_token['data'] == 'expected-closing-tag-but-got-char' and
-                     (self.parser.tags is None or token['data'].lower().strip() not in self.parser.tags))):
+                       self.parser.tags is not None and
+                       token['data'].lower().strip() not in self.parser.tags)):
                     # We've got either a malformed tag or a pseudo-tag or
                     # something that html5lib wants to turn into a malformed
                     # comment which Bleach clean() will drop so we interfere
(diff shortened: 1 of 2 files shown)


g-k:
Thanks! This makes sense. 

Mulling it over @willkg thoughts? or any idea why converting the None to an empty list breaks tests (e.g. https://github.com/mozilla/bleach/compare/parser-tags-nonetype-exception#diff-461f9311bf13103edd11da2cf1e954e3L374)? I'm assuming None is handled as distinct from the empty list, but wanted to confirm.

tests/test_html5lib_shim.py
```diff
@@ -117,6 +117,11 @@ def test_serializer(data, expected):
         {},
         '<a href=\'http://example.com\'\'>',
         '<a href="http://example.com"></a>'
+    ),
+    (
+        {},
+        '</ chars',
+        '&lt;/ chars',
```
willkg: This test case doesn't have anything to do with "invalid-character-in-attribute-name" testing, so I'd add a comment along these lines:

```
# Test that "expected-closing-tag-but-got-char" works when tags is None
```

tests/test_clean.py
```diff
@@ -24,11 +24,11 @@ def test_only_text_is_cleaned():
 
     with pytest.raises(TypeError) as e:
         clean(some_type)
-    assert "argument cannot be of 'type' type" in str(e)
+    assert "argument cannot be of 'type' type" in str(e.value)
```
willkg: The `requirements-dev.txt` file specifically uses `pytest<5`. Updating the code to handle recent versions of pytest should be done in a separate PR.

bleach/html5lib_shim.py
```diff
@@ -256,7 +256,7 @@ def __iter__(self):
                     yield token
 
                 elif ((last_error_token['data'] == 'expected-closing-tag-but-got-char' and
-                     token['data'].lower().strip() not in self.parser.tags)):
+                     (self.parser.tags is None or token['data'].lower().strip() not in self.parser.tags))):
```
willkg: I think this isn't the right fix. If the user explicitly creates a parser that has `tags=None`, then I'm pretty sure it should not look at tags at all and do whatever html5lib does by default.

I think we need to carry that through here, too. So then this should be:

```
self.parser.tags is not None and token['data'].lower().strip() not in self.parser.tags
```

That results in the expected output of the added test being different.

willkg:
Per the docstring here:

https://github.com/mozilla/bleach/blob/master/bleach/html5lib_shim.py#L366-L367

passing in `None` shuts off the special handling for tags that we had to do.

Thus `None` is different than a `[]`. The former skips the special handling the code. The latter says no tags are allowed and either strips or escapes all tags.

tests/test_clean.py
g-k: Yeah, this predates that. So I think we should focus on the second two commits.

bope: Removed this commit after rebase from master.

bleach/html5lib_shim.py
bope: I've updated fix and test as you suggested

g-k:
Landed rebased version in #493. Thank you!

Issue #369, opened 2018-05-01
[Accessibility bug] Semantic whitespace implied by block elements isn't retained properly


JonasT:
Separate block elements imply a visual spacing when seen on screen, which should be retained when removing those elements to keep proper word separation. Bleach doesn't seem to handle this right now:

```
>>> import bleach
>>> html = "<p>Te<b>st</b>!</p><p>Hello</p>"
>>> bleach.clean(html, tags=[], strip=True)
'Test!Hello'
>>>
```

The expected result would be: `'Test! Hello'` (since `<p>` is a block element, and therefore two of them after each other implies a visual line break that is vital for proper readability of the text)

willkg:
Can you make a list of which block elements you want to handle?

JonasT:
https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements#Elements using this or a similar list would be a good idea I think

willkg:
Transcribing that list here:

* address
* article
* aside
* blockquote
* canvas
* dd
* div
* dl
* dt
* fieldset
* figcaption
* figure
* footer
* form
* h1, h2, h3, h4, h5, h6
* header
* hgroup
* hr
* li
* main
* nav
* noscript
* ol
* output
* p
* pre
* section
* table
* tfoot
* ul
* video

Can you take a stab at implementing this? I don't think I'm going to get to this for a while.

JonasT:
I wrote a rich text layouter recently that imports HTML where I got this as a by product, so right now I have no immediate need for this. I just thought it'd be a nice thing to add at some point

jvanasco:
I took a stab at this ticket. The solution is not perfect but allows for readable and accessible text.

While discussion on my approach should continue on the PR I wanted to bring up on this discussion:

The whitespace character should be a NEWLINE and not a SPACE.

    input: <p>Te<b>st</b>!</p><p>Hello</p>
    output:
    - 'Test! Hello'
    + 'Test!\nHello'

The main reason is because more complex use cases, such as lists, become unreadable with a space character -- all the blocks bleed together.

JonasT:
> The whitespace character should be a NEWLINE and not a SPACE.

Yeah you are correct I got that wrong, between block elements there should be a newline since that is also how it is rendered in a browser :+1:

PR #460, opened 2019-07-11
fix for issue #369

jvanasco:
This is an attempt at #369.  I am not particularly happy with it, but I could not figure out a better way to handle this.  The start/end tokens are already collapsed by the time we get to any of the sections I would feel more comfortable handling this in.  More tests are needed, but i'll defer that to others.

Changes:
* `HTML_TAGS__BLOCK_LEVEL` contains the list of block level elements
* `InputStreamWithMemory` now tracks the last emitted token, because...
* `InputStreamWithMemory` emits a SPACE instead of no-space if the following conditions are met:
 1. We have already emitted a token; this guards against a leading space.
 2. We are a START token; this guards against a trailing space.
 3. The last emitted token did not end in a space; this guards against double spaces.


jvanasco:
closing this because I found some edge cases on a larger test suite. will fix and refile.

PR #461, opened 2019-07-12
Fix semantic whitespace (Issue #369)

jvanasco:
After testing with many edge-cases, I decided two things:

1. minimizing whitespace is far too difficult due to the design of this package.  the 'stripping' of tags occurs in a place that can not look backwards enough and can not access the 'content' without a large redesign

2. the whitespace should be a newline, not a space (as suggested in #369). this is because content becomes too confusing to understand with a "space" (such as list elements)


jvanasco:
Closing in favor of a new PR against the current master.

PR #465, opened 2019-08-07
Make the documentation reproducible

lamby:
Whilst working on the [Reproducible Builds effort](https://reproducible-builds.org/) I noticed that python-bleach could not be built reproducibly.

This is because the documentation included a default arguments that was (originally) generated from a `frozenset` type which are iterated over at runtime in a nondeterministic order.

This commit therefore ensures a regular expression is constructed in a deterministic manner, rendering the documentation reproducible.

This bug was originally filed in Debian as [#934120](https://bugs.debian.org/934120).

### DIFF
--- a/bleach/linkifier.py
+++ b/bleach/linkifier.py
@@ -49,7 +49,7 @@ def build_url_re(tlds=TLDS, protocols=html5lib_shim.allowed_protocols):
         (?:[/?][^\s\{{\}}\|\\\^\[\]`<>"]*)?
             # /path/zz (excluding "unsafe" chars from RFC 1738,
             # except for # and ~, which happen in practice)
-        """.format('|'.join(protocols), '|'.join(tlds)),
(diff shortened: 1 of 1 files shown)


jezdez:
@willkg Hey, so I noticed I lost access to this repo, any reason for this?

willkg:
@jezdez I don't work on Bleach anymore, so generally I wouldn't know and you should talk with @g-k. However, I did do a pass a week ago while hanging with @g-k in Portland and dropped admin/owner permissions from anyone who wasn't currently a maintainer. You probably lost it then.

lamby:
Thanks folks :)

jezdez:
> @jezdez I don't work on Bleach anymore, so generally I wouldn't know and you should talk with @g-k. However, I did do a pass a week ago while hanging with @g-k in Portland and dropped admin/owner permissions from anyone who wasn't currently a maintainer. You probably lost it then.

Ack, it would have been nice to be informed about this decision :)

PR #469, opened 2019-09-05
Stop using deprecated Travis container environment

g-k:
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Carries #468 without merging master back into the feature branch.


g-k:
Previously reviewed in #468 (and unchanged since then).

Issue #472, opened 2019-09-05
add example using linkify with custom TLDs to docs


g-k:
refs: #462 #223

PR #473, opened 2019-09-05
docs: Add linkify Linker examples using custom TLDs and protocols

g-k:
fixes: #472 see the linked issues for additional info.

Surfaces the `build_url_re` docstring and adds a couple examples to the docs. One possible downside is if we move away from the `url_re` kwarg then it'll break backwards compatibility, but I think that's OK.

r? @jdufresne or @willkg

### DIFF
--- a/docs/linkify.rst
+++ b/docs/linkify.rst
@@ -349,6 +349,21 @@ domains (TLDs) and URL protocols/schemes:
    'https links <a href="https://example.com/" rel="nofollow">https://example.com/</a>'
 
 
+Specify localized TLDs with and without punycode encoding to handle
+both formats:
+
+.. doctest::
+
+   >>> from bleach.linkifier import Linker, build_url_re
+
+   >>> linker = Linker(url_re=build_url_re(tlds=['рф']))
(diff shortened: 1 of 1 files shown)


g-k:
NB: added a commit for localized TLDs

docs/linkify.rst
```diff
@@ -324,6 +324,48 @@ instance.
    'a b c <a href="http://example.com" rel="nofollow">http://example.com</a> d e f'
 
 
+It includes optional keyword arguments to specify allowed of top-level
```
willkg: I think this should be "to specify allowed top-level".

g-k:
Ah my bad. I pushed some changes carried from #478

g-k:
I can make that a separate PR.

docs/linkify.rst
g-k: Yep! Fixed.

g-k:
`make doctest` from `docs/` passes

> Do the examples get tested when building docs?

I'm not sure if sphinx-build runs it

willkg:
Looks like we did part of the work, but didn't get it running in CI. I threw that together in pull #481.

Issue #464, opened 2019-07-29
`.linkify` vulnerable to punycode attacks


jvanasco:
This was originally filed to the [security bug tracker](https://bugzilla.mozilla.org/show_bug.cgi?id=1566541) but it was decided this is not within the security threat model of the library.

---

Steps to reproduce:

I decided to start work on a PR for the Bleach issue concerning localized domains (https://github.com/mozilla/bleach/issues/368). When building test-cases, I decided to generate one for "punycode attacks" aka [IDN Homograph Attack](https://en.wikipedia.org/wiki/IDN_homograph_attack) when unicode lookalike/homoglyph characters are replaced in a link text.

This allows a malicious attacker to create a link to a lookalike domain name.

For example:

    googĺe.com	xn--googe-95a.com
    wikipediа.org wikipedi\u0430.org

    print linkify("http://googĺe.com")
    print linkify("http://wikipediа.org")

Modern browsers defend against this somewhat by customizing what appears in the display bar once it is clicked; however they do not control (to my knowledge) what happens on the page.

Actual results:

    print linkify("foo http://googĺe.com bar")
    foo <a href="http://googĺe.com" rel="nofollow">http://googĺe.com</a> bar

    print linkify("foo http://wikipediа.org bar")
    foo <a href="http://wikipediа.org" rel="nofollow">http://wikipediа.org</a> bar

Expected results:

Since bleach is a security focused library, I think it should - be default - approach this situation how the strictest of modern browsers do, but allow for other use cases as this approach is very much tailored to "English users" and not the international community. I think a Boolean argument to linkify could be used to provide security by default, and allow non-english alphabet users the ability to opt-out.

Default: the HREF should be be the punycode version

    <a href="http://xn--googe-95a.com" rel="nofollow">http://googĺe.com</a>

Optional Strict: Linkified Text is also punycode

    <a href="http://xn--googe-95a.com" rel="nofollow">http://xn--googe-95a.com</a>

Optional Loose: allow the text as-is and leave everything to the browser

     <a href="http://googĺe.com" rel="nofollow">http://googĺe.com</a>

This functionality could also be implemented in a callback function, which would accept as input two arguments: the raw link and the context ('a' href value, 'a' node/display text,  bare html to be turned into an a tag).

Handling this via a callback is IMHO the best option as it would allow for bleach to abandon the regex of locked-down TLDs that seems to create more issues than it solves, while allowing users to still lock them down if wanted.

willkg:
[...]

PR #476, opened 2019-09-05
docs: add IDN homograph attacks to goals and linkify

g-k:
refs: #464 and discussion in the associated BMO bug

r? @willkg 

If this looks good, I'll also open an issue for adding examples of how one might handle those threats sanely.

### DIFF
--- a/docs/linkify.rst
+++ b/docs/linkify.rst
@@ -26,9 +26,22 @@ having URLs in tag attributes getting linkified.
    You may pass a ``string`` or ``unicode`` object, but Bleach will always
    return ``unicode``.
 
+.. note::
+
+   By default `linkify` **does not** attempt to protect users from bad
+   or deceptive links including:
+
+   * links to malicious or deceptive domains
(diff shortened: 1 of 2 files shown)


docs/linkify.rst
```diff
@@ -26,9 +26,22 @@ having URLs in tag attributes getting linkified.
    You may pass a ``string`` or ``unicode`` object, but Bleach will always
    return ``unicode``.
 
+.. note::
 
-.. autofunction:: bleach.linkify
+   `linkify` **does not** attempt to protect users from bad or
```
willkg: Does it make it clearer to say "by default" here thereby distinguishing Bleach's responsibilities from developer's ability to extend/augment the default behavior using callbacks?

g-k: Yeah good call! Added "by default". We should link to examples / docs / extras functions that handle the homograph attacks once they exist.

g-k:
Thanks Will!

PR #478, opened 2019-09-10
docs: Add linkify example using tlds package

philippbosch:
Not sure if this is the best way to do this but this PR is supposed to amend mozilla/bleach#473


g-k:
This is a nice example! After thinking about this some more I want to avoid having to install extra packages to get doctests to pass and endorsing any particular tld package in the bleach docs.

It'd be good to add the example to the tlds package docs.

PR #481, opened 2019-09-25
Fix tox.ini to run doctests in docs

willkg:
This fixes `tox.ini` to run doctests with the `docs` environment. You can run it manually like this:

```
$ tox -e docs
```

I also tweaked the CI setup to build the docs and run doctests in CI. I think that'll work.


g-k:
Thank you!

PR #486, opened 2019-11-22
Add build_email_re() and use list of TLDs

pc-coholic:
Right now, emails are being parsed through either through the static `EMAIL_RE` or a user-provided regex.

However: parsing emails will however fail for a variety of gTLDs - even if those would be included in the list of `TLDS`, because bleach assumes that any tld cannot be longer than 6 characters.

My first instinct would have been to pass a new `email_re` to the `Linker` and use a `build_email_re()` the same way as the `build_url_re()`.

Since this function did not exist yet, I took the liberty to implement it in this PR - thus allowing the discovery and linking of all TLDs in the same way as for the URL-detection.

The test suite still passes - so I'll assume that this change won't be breaking anything for users opting not to provide any custom list of TLDs.


bleach/linkifier.py
```diff
@@ -59,15 +59,30 @@ def build_url_re(tlds=TLDS, protocols=html5lib_shim.allowed_protocols):
...
+
+    """
+    return re.compile(
+        r"""(?<!//)
+        (([-!#$%%&'*+/=?^_`{}|~0-9A-Z]+
+            (\.[-!#$%%&'*+/=?^_`{}|~0-9A-Z]+)*  # dot-atom
+        |^"([\001-\010\013\014\016-\037!#-\[\]-\177]
+            |\\[\001-\011\013\014\016-\177])*"  # quoted-string
+        )@(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:%s))  # domain
+        """ % '|'.join(tlds),
```
g-k: Having to double the other `%` chars is confusing. Can you change that to a `str.format` like `build_url_re`?

raphaelm: Since there are ``{}`` inside the regex, this would be just as confusing, wouldn't it?

g-k: Yeah, good point. We probably need a comment either way. I'd still like to stick with `str.format` for consistency with the `build_url_re`.

pc-coholic: @g-k Changed it to `str.format` as requested :)

g-k: Thanks!

g-k:
Landed in #496. Thank you!

Issue #425, opened 2018-12-24
Test warnings for python3.7


kitterma:
With python3.7, I get the following warnings from the test suite:

```
=============================== warnings summary ===============================
/usr/lib/python3/dist-packages/html5lib/_trie/_base.py:3
  /usr/lib/python3/dist-packages/html5lib/_trie/_base.py:3: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Mapping

bleach/sanitizer.py:428
  /tmp/buildd/python-bleach-3.0.2/.pybuild/cpython3_3.7_bleach/build/bleach/sanitizer.py:428: DeprecationWarning: invalid escape sequence \s
    "[`\000-\040\177-\240\s]+",

bleach/sanitizer.py:577
  /tmp/buildd/python-bleach-3.0.2/.pybuild/cpython3_3.7_bleach/build/bleach/sanitizer.py:577: DeprecationWarning: invalid escape sequence \s
    style = re.compile('url\s*\(\s*[^\s)]+?\s*\)\s*').sub(' ', style)

bleach/sanitizer.py:592
  /tmp/buildd/python-bleach-3.0.2/.pybuild/cpython3_3.7_bleach/build/bleach/sanitizer.py:592: DeprecationWarning: invalid escape sequence \s
    if not re.match("^\s*([-\w]+\s*:[^:;]*(;\s*|$))*$", style):

bleach/sanitizer.py:596
  /tmp/buildd/python-bleach-3.0.2/.pybuild/cpython3_3.7_bleach/build/bleach/sanitizer.py:596: DeprecationWarning: invalid escape sequence \w
    for prop, value in re.findall('([-\w]+)\s*:\s*([^:;]*)', style):
```

willkg:
What version of Bleach is that? Is that master tip?

willkg:
I checked and all of these except the html5lib one have been fixed in master.

The warning from html5lib was fixed in https://github.com/html5lib/html5lib-python/commit/4f9235752cea29c5a31721440578b430823a1e69 but that hasn't made it into a release, yet. Once that's released, I can update our vendored copy.

brettcannon:
FYI the html5lib issue now raises an `ImportError` in CPython `master` (i.e. 3.9).

fschulze:
[...]

PR #500, opened 2019-12-13
Fix collections ABCs deprecation warning

paxet:
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    from collections import Mapping

### DIFF
--- a/bleach/_vendor/html5lib/_trie/_base.py
+++ b/bleach/_vendor/html5lib/_trie/_base.py
@@ -1,6 +1,11 @@
 from __future__ import absolute_import, division, unicode_literals
 
-from collections import Mapping
+try:
+    # Python 3
+    from collections.abc import Mapping
+except ImportError:
+    # Python 2.7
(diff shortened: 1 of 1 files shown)


g-k:
Thanks! Per https://github.com/mozilla/bleach/issues/425#issuecomment-450687027 we're waiting on an html5lib release with the upstream fix. In the meantime you can suppress the warning like the pypa/twine PR referenced in the above issue.

ankostis:
(my apologies if i misunderstood the situation)
Why is this closed without being merged?
I still see this warning in the package currently in PyPi, v3.1.5.
#533 (merged) seems not to solve the issue, just hides the warning during testing, is this on purpose?

paxet:
Thank you so much, finally closed by #548 and published in [v3.2.0](https://github.com/mozilla/bleach/releases/tag/v3.2.0)

PR #502, opened 2019-12-17
only install six for python2 versions

shaneikennedy:
I noticed that for my application running python 3.6 that installing bleach was still installing six, which is not needed. This PR modifies the setup arguments to match the [setuptools docs](https://setuptools.readthedocs.io/en/latest/setuptools.html#id17) for declaring platform specific dependencies.

Let me know what you think and if there's anything else I can do here :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ tests_require = [
 ]
 
 install_requires = [
-    'six>=1.9.0',
+    'six>=1.9.0;python_version<"3"',
     # html5lib requirements
     'webencodings',
 ]


g-k:
Thanks! This would be good, but I don't think we can land it. Bleach vendors html5lib which depends on six so we'd need html5lib to drop six first. CI passes because the dev tools install six.

Without the dev tools install I'm seeing:

```console
~/bleach - [drop-six●] » python -m venv ./venv    
~/bleach - [drop-six●] » source venv/bin/activate
(venv) ~/bleach - [drop-six●] » python setup.py develop
...
(venv) ~/bleach - [drop-six●] » python
Python 3.8.0 (default, Nov 21 2019, 17:41:43)
[GCC 7.4.0] on linux
>>> import bleach
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/gguthe/bleach/bleach/__init__.py", line 7, in <module>
    from bleach.linkifier import (
  File "/home/gguthe/bleach/bleach/linkifier.py", line 3, in <module>
    import six
ModuleNotFoundError: No module named 'six'
>>>
```

PR #506, opened 2019-12-27
Adding Travis-Ci Support For Arm64

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 job in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,30 +5,64 @@ cache:
   directories:
   - "~/.cache/pip"
 
-python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
-  - "3.6"
-  - "3.7"
-  - "3.8"
-  - "pypy2.7-6.0"
-  - "pypy3.5-6.0"
+matrix:
+  include:
+    - arch: arm64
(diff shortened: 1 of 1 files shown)


ossdev07:
Ping @g-k can you have a look at this PR.

g-k:
Hi @ossdev07, thanks for the PR! What's the benefit of testing on ARM64? Have you noticed any regressions that only occur on ARM?

AFAIK bleach doesn't have any native dependencies.

rhenwood-arm:
Hi @g-k , I have been working with @ossdev07 on a down-stream project on AArch64 (ARM64) that has bleach as a dependency. As you point out, bleach is a pure python project with a straightforward Travis-CI build job and that doesn't have a binary release. In this case, there is very little benefit in testing on alternative architectures, and this PR can probably be held in reserve until it is needed.

g-k:
Ok thanks! Closing per https://github.com/mozilla/bleach/pull/506#issuecomment-585419723 we would like to know about architecture specific regressions, but will rely on downstream consumers to report them.

Issue #503, opened 2019-12-20
Test suite fails with Python 3.8.1


Ma27:
This happens since `urllib.parse` parses URLs like `path:80` as `scheme=path,path=80` (https://bugs.python.org/issue27657).

Which causes errors like this:

```
def test_uri_value_allowed_protocols(data, kwargs, expected):
>       assert clean(data, **kwargs) == expected
E       assert '<a>valid</a>' == '<a href="192...00">valid</a>'
E         - <a>valid</a>
E         + <a href="192.168.100.100:8000">valid</a>

tests/test_clean.py:676: AssertionError
```

Currently preparing a patch for this :)

hugovk:
@Ma27 Hi, any progress with the patch? Thanks!

Ma27:
Huh, I thought I provided a status update here as well, sorry!

Unfortunately I'm not sure if I actually want to work around this, the code already depends on some heuristics and I'm afraid that patching around in there will cause some mean regressions.

I still hope that the change gets reverted again: https://github.com/python/cpython/pull/16839#issuecomment-570758153

kitterma:
I don't see much evidence of upstream python3.8 activity to revert.  Even if they do, the same problem will occur once 3.9 is out, so it's got to be solved sooner or later.

Ma27:
>  Even if they do, the same problem will occur once 3.9 is out, so it's got to be solved sooner or later.

Agreed. While taking a look at this a few months ago I realized though that sanitizing has more culprits than I anticipated, so unless the change is reverted, someone else should fix the library accordingly.

PR #514, opened 2020-01-28
Test and lint on GitHub Actions

hugovk:
This replicates the current Travis CI setup, except:

* Doesn't test EOL Python 2.7 and 3.4. These can be added if you like.
* Doesn't test Python 3.7 and 3.8, as they're currently failing due to #503:
   * https://github.com/hugovk/bleach/runs/412523804?check_suite_focus=true
* lint/vendorverify/docs are run on 3.8

Tests are run on all available operating systems:
* `ubuntu-18.04`, `ubuntu-16.04`, `macos-latest`, `windows-latest`
* If there's no need to test on both Ubuntus, `ubuntu-latest` can be used as well

I made one workflow config file for testing, and another for linting.

They're mostly identical, due to caching boilerplate. They could be combined into a single config file, but the matrix definition might be a bit complicated and difficult to maintain. (I've also proposed a change to pip to allow the cache boilerplate to be simplified in the future.) But if you like, I can check how it'd look.

Here's how a test run looks:

* https://github.com/hugovk/bleach/runs/412558440?check_suite_focus=true

And lint:

* https://github.com/hugovk/bleach/runs/412558503?check_suite_focus=true

---

I've not tried it yet (on GHA, I have on Travis), but here's one way to deploy to PyPI:

* https://github.com/pypa/gh-action-pypi-publish


g-k:
This is great! I need to get in touch with a GH org owner about enabling actions for the mozilla org or find out if we can transfer the repo somewhere.

hugovk:
Unlike other CIs, I don't think GitHub Actions needs explicitly enabling, just merging to get the config files in the repo for the first time.

hugovk:
Updated to simplify config with newer GHA releases.

---

> * Doesn't test Python 3.7 and 3.8, as they're currently failing due to #503:
>   * [hugovk/bleach/runs/412523804?check_suite_focus=true](https://github.com/hugovk/bleach/runs/412523804?check_suite_focus=true)

Undone this, and added 3.7 and 3.8 now they pass.

---

The vendorify/lint now fails on GHA (https://github.com/hugovk/bleach/runs/1112845665?check_suite_focus=true), but it's also failing on Travis CI on `master` (https://travis-ci.com/github/hugovk/bleach/builds/184230513) and this PR.

g-k:
Awesome! This will also fix #570.

We recently got actions approved, and since this is using first party actions and bleach isn't a sensitive repo we should be fine in terms of https://wiki.mozilla.org/GitHub#How_do_I_hook_up_a_new_GitHub_Actions_or_3rd_party_application_to_a_repository_in_the_mozilla_org.3F

g-k (approved):

PR #521, opened 2020-03-12
import ABCs from collections.abc for Python 3.9

aaronchall:


mgorny:
It would be better to stop vendoring very old version of html5lib and instead use upstream sources where this was fixed back in 2018.

graingert:
`from six.moves import collections_abc` btw

domdfcoding:
The latest version of html5lib is now 1.1 (released Jun 23, 2020) , which looks to have this fix included.

PR #531, opened 2020-04-28
Replace setuptools dependency with packaging.

benjaminp:
bleach had a (undeclared) dependency on setuptools via the import of pkg_resources in __init__.py. pkg_resources is mostly deprecated, and its use should be avoided. For bleach's usage, we can directly employ the packaging library.


g-k:
Thanks!

PR #533, opened 2020-05-13
Update html5lib to Py3 Mapping import

mdinsmore:
The bleach project uses html5lib 1.0.1 - the latest version on pypi - however this generates a deprecation warning for the Mapping import in bleach/_vendor/html5lib/_trie/_base.py

This pull request updates the import to the python 3.3+ form while maintaining backwards compatibility as per https://github.com/html5lib/html5lib-python/blob/master/html5lib/_trie/_base.py#L3

### DIFF
--- a/bleach/_vendor/html5lib/_trie/_base.py
+++ b/bleach/_vendor/html5lib/_trie/_base.py
@@ -1,6 +1,9 @@
 from __future__ import absolute_import, division, unicode_literals
 
-from collections import Mapping
+try:
+    from collections.abc import Mapping
+except ImportError:  # Python 2.7
+    from collections import Mapping
 
 
 class Trie(Mapping):


domdfcoding:
The latest version of html5lib is now 1.1, which looks to have the fix you propose included. It might, therefore, be more prudent to vendor the latest version into bleach. The only breaking change is dropping support for Python 3.3 and 3.4, but bleach doesn't claim to support those versions anyway.

There's also another open PR, #521, which fixes the imports but drops Python 2 support entirely.

mdinsmore:
> The latest version of html5lib is now 1.1, which looks to have the fix you propose included. It might, therefore, be more prudent to vendor the latest version into bleach. The only breaking change is dropping support for Python 3.3 and 3.4, but bleach doesn't claim to support those versions anyway.
> 
> There's also another open PR, #521, which fixes the imports but drops Python 2 support entirely.

Cool glad there's a new version of html5lib - when I submitted the request they had updated the code on github but not released a new version - looking forward to the new version of bleach - thanks for your work on it.

ovillemain (approved):

Issue #520, opened 2020-03-12
Drop support for EOL Python versions <3.6


g-k:
Kinda blocked on html5lib switching to Python 3, but this would:

* let us drop the dependency on `six` (pending html5lib doing the same)
* add Python 3 type annotations (outside comments) 
* use other Python 3 features
* reduce compat and testing burden

TODO: 

- [ ] ~~check Python 2/2.7 usage stats https://pypistats.org/packages/bleach~~
- [ ] ~~see if major library consumers in the ecosystem would be impacted~~

Given that CPython 2.7 support is ending I think it's safe to drop Python 2 support with the next major release.

jvanasco:
What about announcing that `4.0` will drop Python2 support, so people can pin dependencies for auto upgrades now?

g-k:
Absolutely, do you have any suggestions on where to announce that?

g-k:
We should also drop support for unsupported python 3 versions too currently everything <3.6 https://en.wikipedia.org/wiki/History_of_Python#Version_3

PR #535, opened 2020-05-15
Add style attr sanitizer

g-k:
As discussed in #248, we'd like to move away from the regex based CSS sanitizer but don't necessarily have a maintained CSS parser with Python 2 support to switch to (also some users don't necessarily want full CSS parsing).

Changes in this PR:

* add a `css_style_attr_sanitizer` param/arg to bleach `Cleaner` and `BleachSanitizerFilter` to make it possible to override `sanitize_css` without changing `Cleaner.clean`
* update the arbitrary style goal

We'd then cut a major release that:

* drops python 2 support #520
* replaces old regex based stuff with tinycss2 #248

r? @jdufresne 

cc @willkg @peterbe

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,21 @@
 Bleach changes
 ==============
 
+Version 3.2.0.dev0 (May 15th, 2020)
+-----------------------------------
+
+**Security fixes**
+
+None
+
+**Features**
+
+* Add ``css_style_attr_sanitizer`` argument to ``bleach.sanitizer.Cleaner``
+
+**Bug fixes**
+
+None
+
 Version 3.1.5 (April 29th, 2020)
 --------------------------------
 
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -18,9 +18,9 @@ from bleach.sanitizer import (
 
 
 # yyyymmdd
-__releasedate__ = '20200429'
+__releasedate__ = '20200515'
 # x.y.z or x.y.z.dev0 -- semver
-__version__ = '3.1.5'
+__version__ = '3.2.0.dev0'
 VERSION = packaging.version.Version(__version__)
 
 
--- a/docs/clean.rst
+++ b/docs/clean.rst
@@ -395,3 +395,32 @@ use an html5lib filter.
 
 
 .. versionadded:: 2.0
+
+
+Using different CSS parser and sanitizer with `css_style_attr_sanitizer`
+------------------------------------------------------------------------
+
+The argument `css_style_attr_sanitizer` can ``bleach.sanitizer.Cleaner`` be used
+to customize CSS parsing and sanitization. For example, define a new sanitization
+function:
+
+   >>> from bleach import Cleaner
+
+   >>> def sanitize_css(style):
+   ...     print('ignoring style attribute value:', repr(style))
+   ...     return ''
+
+   Create a new :py:class:`bleach.sanitizer.Cleaner` using it:
+
+   >>> cleaner = Cleaner(
+   ...     tags=['p'],
+   ...     attributes=['style'],
+   ...     styles=['not used'],
+   ...     css_style_attr_sanitizer=sanitize_css,
+   ... )
+
+   >>> cleaner.clean('<p style="cursor: -moz-grab;">bar</p>')
+   ignoring style attribute value: 'cursor: -moz-grab;'
+   '<p style="">bar</p>'
+
+.. versionadded:: 3.2.0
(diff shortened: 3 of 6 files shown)


CHANGES
```diff
@@ -1,6 +1,21 @@
 Bleach changes
 ==============
 
+Version 3.2.0.dev0 (May 15th, 2020)
```
willkg, line 4: I only set the date here if it's been released. Otherwise it gets confusing in the documentation as to what was released and what's in the master branch that hasn't been released, yet.

So for this, I'd use "In development" or something along those lines.

bleach/__init__.py
```diff
@@ -18,9 +18,9 @@
 
 
 # yyyymmdd
-__releasedate__ = '20200429'
+__releasedate__ = '20200515'
```
willkg, line 21: I only set the date here for releases. When it's in development, I leave this as the empty string.

docs/clean.rst
```diff
@@ -395,3 +395,32 @@ use an html5lib filter.
 
 
 .. versionadded:: 2.0
+
+
+Using different CSS parser and sanitizer with `css_style_attr_sanitizer`
+------------------------------------------------------------------------
+
+The argument `css_style_attr_sanitizer` can ``bleach.sanitizer.Cleaner`` be used
```
willkg, line 403: This sentence doesn't parse in my head. Do you mean something like this?

> The argument css_style_attr_sanitizer in bleach.sanitizer.Cleaner can be used ..

docs/clean.rst
```diff
@@ -395,3 +395,32 @@ use an html5lib filter.
...
+to customize CSS parsing and sanitization. For example, define a new sanitization
+function:
+
+   >>> from bleach import Cleaner
+
+   >>> def sanitize_css(style):
```
willkg, line 409: I'd rename this to `my_sanitize_css` to make it clearer it's the overridden one rather than the Cleaner instance method.

docs/goals.rst
```diff
@@ -133,16 +133,32 @@ Bleach to make it safe and then do your own transform afterwards.
...
+
+* `XSS in legacy browsers <https://blog.innerht.ml/cascading-style-scripting/>`_ using properties like ``-o-link`` and
+  ``-moz-binding``, or functions like ``expression()`` and ``url()``
+
+
+Bleach doesn't know how the surrounding page is styled or what a
+site's authors or users consider customization instead of defacement.
+
+Consequently, bleach requires you to opt in if you want your users to
+be able to change nearly anything in a ``style`` attribute.
```
willkg, line 154: This is much more useful than what we had before. :clap:

willkg:
I like this, but I wonder if the ongoing "Bleach does it right by default but then lets users do whatever with warnings of dire but vague consequences" architecture changes we're making should also be accompanied by something that can help users know when their overrides create bad situations.

Could we create a test harness that people could pass a cleaner to and it helps them figure out if they've got vulnerabilities?

Maybe we can base it on the OWASP-derived test cases we've got now? (https://github.com/mozilla/bleach/tree/master/tests/data) Maybe we can base it on the website harness we've got? Maybe there's a page safety site out there already we could use?

I'm just wondering out loud. There's no way users on average understand the problem domain enough to know a bad decision when they're overriding Bleach thing.

g-k:
Thanks for the review willkg!

> I wonder if the ongoing "Bleach does it right by default but then lets
users do whatever with warnings of dire but vague consequences"
architecture changes we're making should also be accompanied by something
that can help users know when their overrides create bad situations.

> Could we create a test harness that people could pass a cleaner to and it
helps them figure out if they've got vulnerabilities?

> Maybe we can base it on the OWASP-derived test cases we've got now? (
https://github.com/mozilla/bleach/tree/master/tests/data) Maybe we can base
it on the website harness we've got? Maybe there's a page safety site out
there already we could use?

> I'm just wondering out loud. There's no way users on average understand
the problem domain enough to know a bad decision when they're overriding
Bleach thing.

Yeah, library consumers shouldn't need to know details of what threats
bleach is trying to protect against. I think there are two things we can
do:

1. take a cue from https://cryptography.io/en/latest/#layout and split the
API into a safe and potentially a more stable "recipes" layer (just
bleach.clean, bleach.linkify, and maybe some other top-level functions
taking a limited subset of args) and a more dangerous "hazmat" layer with
all the knobs and buttons to tweak and override things. Security bugs
against the recipes layer would be higher severity.

2. create an evaluator (like various CSP ones) or config generator (like
the Moz SSL config generator) for people using the recipes or hazmat layer
for consumers to provide additional context (where they're using bleach
output, who's entering the data and how much they trust them, what other
controls are in place, etc.) and to know whether their bleach config is
safe.

Issue #557, opened 2020-09-17
bug: DeprecationWarning: html5lib's sanitizer is deprecated


pllim:
**Describe the bug**

Hello. This deprecation has come up in our project's CI as such ([full log](https://travis-ci.org/github/astropy/astropy/jobs/727949664)):
```
.../astropy/utils/xml/writer.py:195: in <lambda>
    self.xml_escape_cdata = lambda x: bleach.clean(x, **clean_kwargs)
.../bleach/__init__.py:84: in clean
    return cleaner.clean(text)
.../bleach/sanitizer.py:175: in clean
    filtered = BleachSanitizerFilter(
.../bleach/sanitizer.py:273: in __init__
    return super(BleachSanitizerFilter, self).__init__(source, **kwargs)
...
E       DeprecationWarning: html5lib's sanitizer is deprecated;
see https://github.com/html5lib/html5lib-python/issues/443 and
please let us know if Bleach is unsuitable for your needs
```
As the traceback says, we are already using Bleach (3.2.0).

I asked over at https://github.com/html5lib/html5lib-python/issues/443#issuecomment-694211077 and it was suggested that I open a bug report here. Thank you!

**python and bleach versions (please complete the following information):**

 - Python Version: 3.8.1
 - Bleach Version: 3.2.0

**To Reproduce**

Steps to reproduce the behavior: Is the traceback sufficient?

**Expected behavior**

No deprecation warning, because I am not sure if there is anything I can do about it, except ignoring the warning.

**Additional context**

Our CI turns deprecation warnings into failures.

g-k:
Hi, thanks for the report!

The traceback is sufficient. I'll add some tests that warnings aren't propagated.

PR #558, opened 2020-09-17
Fix 557 suppress html5lib sanitizer deprecation warnings

g-k:
fixes: #557 

r? @willkg if you get a minute

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -12,26 +12,51 @@ import warnings
 
 import six
 
-# filter out html5lib deprecation warnings to use bleach
-warnings.simplefilter("ignore", category=DeprecationWarning)
+# ignore html5lib deprecation warnings to use bleach; we are bleach
+# apply before we import submodules that import html5lib
+warnings.filterwarnings(
+    "ignore",
+    message="html5lib's sanitizer is deprecated",
+    category=DeprecationWarning,
+    module="bleach._vendor.html5lib",
+)
 
-from bleach._vendor.html5lib import (
+from bleach._vendor.html5lib import (  # noqa: E402 module level import not at top of file
     HTMLParser,
     getTreeWalker,
 )
-from bleach._vendor.html5lib import constants
-from bleach._vendor.html5lib.constants import (
+from bleach._vendor.html5lib import (
+    constants,
+)  # noqa: E402 module level import not at top of file
+from bleach._vendor.html5lib.constants import (  # noqa: E402 module level import not at top of file
     namespaces,
     prefixes,
 )
-from bleach._vendor.html5lib.constants import _ReparseException as ReparseException
-from bleach._vendor.html5lib.filters.base import Filter
-from bleach._vendor.html5lib.filters.sanitizer import allowed_protocols
-from bleach._vendor.html5lib.filters.sanitizer import Filter as SanitizerFilter
-from bleach._vendor.html5lib._inputstream import HTMLInputStream
-from bleach._vendor.html5lib.serializer import HTMLSerializer
-from bleach._vendor.html5lib._tokenizer import attributeMap, HTMLTokenizer
-from bleach._vendor.html5lib._trie import Trie
+from bleach._vendor.html5lib.constants import (
+    _ReparseException as ReparseException,
+)  # noqa: E402 module level import not at top of file
+from bleach._vendor.html5lib.filters.base import (
+    Filter,
+)  # noqa: E402 module level import not at top of file
+from bleach._vendor.html5lib.filters.sanitizer import (
+    allowed_protocols,
+)  # noqa: E402 module level import not at top of file
+from bleach._vendor.html5lib.filters.sanitizer import (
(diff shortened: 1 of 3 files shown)


bleach/__init__.py
```diff
@@ -1,14 +1,24 @@
...
+# ignore html5lib deprecation warnings to use bleach; we are bleach
+# apply before we import submodules that import html5lib
+warnings.filterwarnings(
+    "ignore",
+    message="html5lib's sanitizer is deprecated",
+    category=DeprecationWarning,
+    module="bleach._vendor.html5lib",
+)
+
+from bleach.linkifier import (  # noqa: E402 module level import not at top of file
```
willkg, line 17: E402 is so annoying. I can't think of a case where it's helped me. Might be worth excluding in the configuration.

willkg:
It looks like this covers two cases:

1. `import bleach`
2. `from bleach import sanitizer`

Is it possible to put the warning-ignoring code at the top of `bleach/html5lib_shim.py` instead? That way all the html5lib-related things are in the same place which simplifies maintenance.

g-k:
I moved the warning ignoring code from `__init__.py` to `bleach/html5lib_shim.py` which covers the warning in html5lib init https://github.com/mozilla/bleach/blob/b54c78e084b41be8fda8483f72740710377edd26/bleach/_vendor/html5lib/filters/sanitizer.py#L29

There's another in the `html5lib.filters.sanitizer::Filter.__init__` https://github.com/mozilla/bleach/blob/b54c78e084b41be8fda8483f72740710377edd26/bleach/_vendor/html5lib/filters/sanitizer.py#L771 

`from bleach import clean; clean('foo')` raises the warning even when the init warnings filter is in place, so I think both are necessary

bleach/html5lib_shim.py
```diff
@@ -12,26 +12,51 @@
...
 
-# filter out html5lib deprecation warnings to use bleach
-warnings.simplefilter("ignore", category=DeprecationWarning)
+# ignore html5lib deprecation warnings to use bleach; we are bleach
+# apply before we import submodules that import html5lib
+warnings.filterwarnings(
+    "ignore",
+    message="html5lib's sanitizer is deprecated",
+    category=DeprecationWarning,
+    module="bleach._vendor.html5lib",
```
willkg, line 21: If you nix this `module=` bit here, then will this apply to the module and the Sanitizer `__init__`?

g-k:
ok so, the `test_clean.py` failure is from:

```
» PYTHONDONTWRITEBYTECODE=1 pytest --cache-clear -x -v tests/test_clean.py
========================================================================== test session starts ==========================================================================
platform linux -- Python 3.8.2, pytest-6.0.2, py-1.9.0, pluggy-0.13.1 -- /home/gguthe/bleach/venv/bin/python
cachedir: .pytest_cache                                        
rootdir: /home/gguthe/bleach, configfile: setup.cfg        
plugins: wholenodeid-0.2                                                                                                                                                 collected 183 items                                                                                                                                                     
                                                             
tests/test_clean.py::test_clean_idempotent FAILED                                                                                                                 [  0%]
                                                                                                                                                                         
=============================================================================== FAILURES ================================================================================
______________________________________________________________ tests/test_clean.py::test_clean_idempotent _______________________________________________________________
                                                                                                                                                                         
    def test_clean_idempotent():                                               
        """Make sure that applying the filter twice doesn't change anything."""
        dirty = "<span>invalid & </span> < extra http://link.com<em>"
>       assert clean(clean(dirty)) == clean(dirty)

tests/test_clean.py:16:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
bleach/__init__.py:90: in clean
    return cleaner.clean(text)
bleach/sanitizer.py:177: in clean
    filtered = BleachSanitizerFilter(
bleach/sanitizer.py:288: in __init__
    return super(BleachSanitizerFilter, self).__init__(source, **kwargs)
```

bleach/html5lib_shim.py
g-k: no that still fails

g-k:
however, a separate test with:

```python
    from bleach import clean
    # from bleach.html5lib_shim import Filter                                                                                                                            
    # from bleach.sanitizer import Cleaner                                                                                                                               
    # from bleach._vendor.html5lib.constants import rcdataElements                                                                                                       
    clean('foo')
```

does not fail

pllim:
Hello. Thank you for the quick fix! In which release should I expect this patch to be in?

g-k:
It's out in 3.2.1 https://pypi.org/project/bleach/3.2.1/

Thanks for reporting the bug!

pllim:
Amazing. Thank you!

PR #564, opened 2020-11-05
Fix 520 drop python2

### LINKED ISSUE #520: Drop support for EOL Python versions <3.6

g-k:
Fix #520

Changes:

* update CI and tox configs
* update package `setup.py` metadata and `CHANGES`
* replace `six` in non-vendored code

This doesn't:

*  drop `six` since html5lib needs it
* add type annotations (TODO: later, but we can use the non-comment syntax now)
* change code to take advantage of Python 3 features
* land any other changes planned for the 4.x milestone (thinking add those as time permits and increment the major version number)

I'm planning to squash everything down into 1 to 3 commits if it looks OK.

f? @willkg

### DIFF
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
 
-from __future__ import unicode_literals
-
 import packaging.version
 
 from bleach.linkifier import (
--- a/bleach/__init__.py
+++ b/bleach/__init__.py
@@ -18,9 +18,9 @@ from bleach.sanitizer import (
 
 
 # yyyymmdd
-__releasedate__ = "20200918"
+__releasedate__ = ""
 # x.y.z or x.y.z.dev0 -- semver
-__version__ = "3.2.1"
+__version__ = "4.0.0.dev0"
 VERSION = packaging.version.Version(__version__)
 
 
--- a/bleach/linkifier.py
+++ b/bleach/linkifier.py
@@ -1,6 +1,5 @@
 from __future__ import unicode_literals
 import re
-import six
 
 from bleach import callbacks as linkify_callbacks
 from bleach import html5lib_shim
@@ -171,7 +170,7 @@ class Linker(object):
         :raises TypeError: if ``text`` is not a text type
 
         """
-        if not isinstance(text, six.string_types):
+        if not isinstance(text, str):
             raise TypeError("argument must be of text type")
 
         text = force_unicode(text)
--- a/bleach/linkifier.py
+++ b/bleach/linkifier.py
@@ -1,4 +1,3 @@
-from __future__ import unicode_literals
 import re
 
 from bleach import callbacks as linkify_callbacks
--- a/bleach/linkifier.py
+++ b/bleach/linkifier.py
@@ -3,7 +3,7 @@ import re
 
 from bleach import callbacks as linkify_callbacks
 from bleach import html5lib_shim
-from bleach.utils import alphabetize_attributes, force_unicode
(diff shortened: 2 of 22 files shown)


willkg:
A 4.x milestone. That's phenomenal!

I'll try to review this tomorrow. If there's anything you want me to pay particular attention to, let me know.

bleach/__init__.py
```diff
@@ -18,9 +16,9 @@
 
 
 # yyyymmdd
-__releasedate__ = "20200918"
+__releasedate__ = ""
 # x.y.z or x.y.z.dev0 -- semver
-__version__ = "3.2.1"
+__version__ = "4.0.0.dev0"
```
g-k, line 21: should we release a beta first?

bleach/linkifier.py
```diff
@@ -171,11 +169,9 @@ def linkify(self, text):
         :raises TypeError: if ``text`` is not a text type
 
         """
-        if not isinstance(text, six.string_types):
+        if not isinstance(text, str):
             raise TypeError("argument must be of text type")
 
-        text = force_unicode(text)
```
g-k, line 177: `force_unicode` checks text it coerces is valid utf-8, but the encoding arg for `str` uses `sys.getdefaultencoding` so maybe we should validate that.

g-k:
Awesome! No rush on the taking a look. There's a lot going on.

Left a couple comments on things I wasn't sure of.

I don't think I did anything crazy (mainly swapping `six.` variants for the Python 3 specific replacement), but if you can sanity check that I didn't miss anything that'd be good.

bleach/__init__.py
willkg: I think the nicety of an alpha/beta/rc comes only when you've got people lined up to use/test it that are representative of Bleach users. If you don't, then you spend a bunch of time doing the alpha/beta/rc, wait a bunch of time, and then it's not clear whether that extra effort bought you anything or reduced the chances of a follow-up fixit release.

In the past, I've kept tabs on bugs and done follow-up fixit releases. I can usually do that pretty quickly.

If it were me, I'd release on a Monday with a few following days that I could turn around a fixit release and not go the alpha/beta/rc route.

bleach/linkifier.py
willkg: Oh, interesting! Nice catch!

willkg: I don't think we need to worry about this. If the `text` value isn't a str, then we don't need to decode it so we don't need to validate encodings.

I think what you did is right and all we need to do is remove the `force_unicode` call.

setup.py
```diff
@@ -45,7 +45,7 @@ def get_version():
     include_package_data=True,
     package_data={'': ['README.rst']},
     zip_safe=False,
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
```
willkg, line 48: Yay!

willkg:
This looked fine to me. I didn't look at the code that's not covered in this PR, but it's probably fine and if there are other things we could clean up post-dropping-Python-2, we can do them later.

Issue #566, opened 2020-11-24
feature: Allow to preserve order of attributes


askoretskiy:
At the moment `bleach.clean` sorts attributes.

There are several use cases when it is more desired to preserve the order.

Imagine following HTML:

```
<a href="https:/ya.ru" rel="noopener noreferrer" target="_blank" style="font-size: 24px;">Yandex</a>
``` 

The result of `bleach.clean` would be different than the input:

```
>>> import bleach
>>> text = """<a href="https:/ya.ru" rel="noopener noreferrer" target="_blank" style="font-size: 24px;">Yandex</a>"""
>>> bleach.clean(text, attributes=["href", "rel", "target", "style"], styles=["font-size"])
'<a href="https:/ya.ru" rel="noopener noreferrer" style="font-size: 24px;" target="_blank">Yandex</a>'
>>> text
'<a href="https:/ya.ru" rel="noopener noreferrer" target="_blank" style="font-size: 24px;">Yandex</a>'
```

Sometimes it is important to preserve the order of the attributes. For example, to check if the clean version of the HTML is the same as the original.

Technically, attributes are sorted by function `bleach.utils.alphabetize_attributes`.

The reason for this function was to keep stable order of the attributes. Internally attributes are stored as `dict` that prior to Python3.5 were storing elements in "random" order. Starting from Python3.5 all the dicts preserve the order of elements, so they are stable by itself.

I suggest to add a parameter `sort_attributes=True` to `BleachSanitizerFilter` and if it is set to `False` -- then do not sort attributes on fly. I would keep `True` as default value so that behaviour stays the same as now.

askoretskiy:
I could prepare MR if needed

g-k:
Hi, thanks for the PR and digging into this!

Bleach doesn't try to prettify output, so it makes sense to have it default to not sorting attrs. I'd rather not add more flags though, so we can change the behavior in the next major release.

jvanasco:
Before this change rolls out, please release a warning or some sort of message well in advance, as this is likely to break a lot of test suites and probably some production uses.

IMHO, it would be more proper to keep the current behavior and let uses opt-in.  This could even be done with a module or environment variable.

I would have LOVED this feature 8 years ago.  I am eager to adopt it when it is hits, however this is likely to break most of our test suites.

g-k:
I'm not planning to add this to bleach 3.x. I want to avoid adding more flags to `.clean` and retain support for the Python versions 3.x currently supports.

[...]

PR #567, opened 2020-11-25
Allow to preserve order of HTML attributes

askoretskiy:
This would allow to preserve order of HTML attributes. See https://github.com/mozilla/bleach/issues/566 for more details.


g-k:
Closing since I'd like to make this the default in 4.0.0 (as mentioned in https://github.com/mozilla/bleach/issues/566#issuecomment-754778993)

Added you as a contributor in #576.

PR #571, opened 2021-01-05
ci: remove travis

g-k:
fix: #570 (now that hugovk did the hard part)

Also, I didn't enable actions for the repo before landing #514, so this will kick off a job.

Assuming these pass I'll update the required CI checks in branch protection.


g-k (commented):
:eyes:

PR #579, opened 2021-01-26
Validate convert entity code points

g-k:
Fix `convert_entity` exceptions for certain inputs

r? @willkg

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -459,9 +459,22 @@ def convert_entity(value):
     if value[0] == "#":
         if len(value) < 2:
             return None
+
         if value[1] in ("x", "X"):
-            return six.unichr(int(value[2:], 16))
-        return six.unichr(int(value[1:], 10))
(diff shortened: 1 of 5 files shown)


bleach/html5lib_shim.py
```diff
@@ -459,9 +459,22 @@ def convert_entity(value):
     if value[0] == "#":
         if len(value) < 2:
             return None
+
         if value[1] in ("x", "X"):
-            return six.unichr(int(value[2:], 16))
-        return six.unichr(int(value[1:], 10))
+            # hex-encoded code point
+            prefix, base = value[2:], 16
```
willkg, line 465: I think I don't like the word "prefix" here. I was thinking it was the `#x` part, but it's not--it's the value as a string that's going to be converted later.

I don't want to reuse "value", but maybe `num_as_string`? Or `int_as_string`? Something like that?

g-k: yeah good point renamed to `int_as_string`

PR #600, opened 2021-07-14
Fix attribute name in linkify docs.

CheesyFeet:
Think this should be 'title' instead of 'rel' if it refers to the example above?


g-k (approved):
Nice catch! Thank you!

PR #603, opened 2021-07-23
Fix semantic whitespace (Issue #369)

jvanasco:
block elements are tracked and a newline is inserted when they are stripped.

new tests are included.

See Issue #369 

This is a simplified rewrite of #461 against the current main


jvanasco:
This is agains the 3.x branch; for 4.x see #642

PR #612, opened 2021-08-25
Next major release

g-k:


willkg:
I didn't understand what this PR was or that it was for pulling together all the things for the next major release. I don't understand this workflow, so I'm going to switch back to the "regular one" where we land PRs against the main branch and develop the next release there. That's easier to reason about (for me) plus I think it's easier for contributors to contribute to because GitHub defaults to basing PRs against the default branch especially where we don't have any documentation to do otherwise.

I think I want to land type annotations in the 5.0.0 release. I think we should pull that work into a new branch.

g-k:
wfm we haven't needed a point release against 4.x in awhile.

PR #625, opened 2021-12-30
Add notes on migrating from html5lib sanitization

### LINKED ISSUE #624: How to be compatible with html5lib's sanitization?
I've been using html5lib's native sanitization, but I'm migrating to Bleach, as recommended in https://github.com/html5lib/html5lib-python/blob/master/CHANGES.rst#11. I'm replacing my previous code:
```py
html5lib.serialize(html5parser.HTMLParser().parseFragment("<u>hi</u>"), sanitize=True)
# -> <u>hi</u>
```
with this:
```py
bleach.clean("<u>hi</u>")
# -> &lt;u&gt;hi&lt;/u&gt;
```

Notice how Bleach escapes my `<u>hi</u>` because the `u`-tag isn't on its [very conservative tokenizer allow-list](https://github.com/mozilla/bleach/blob/a362171055af36996dc1cedb64c6b3f68359db27/bleach/sanitizer.py#L13). I am aware that I could specify my own list of allowed tags, like so:
```py
bleach.clean("<u>hi</u>", tags=MY_LIST_THAT_INCLUDES_U)
# -> <u>hi</u>
```
but how would I construct `MY_LIST_THAT_INCLUDES_U` such that it's backward-compatible with what html5lib's sanitizer allows (allowed me until now)? I think I'd need to get at the [larger tag set](https://github.com/mozilla/bleach/blob/main/bleach/_vendor/html5lib/filters/sanitizer.py#L31) in the vendored html5lib sanitizer. Is that what I'm supposed to do? Is there a neater way to achieve that? Has this come up as a use case previously?

sferencik:
fixes #624

### DIFF
--- a/docs/migrating-from-html5lib-sanitizer.rst
+++ b/docs/migrating-from-html5lib-sanitizer.rst
@@ -1,0 +1,70 @@
+.. highlight:: python
+
+=====================================
+Migrating from the html5lib sanitizer
+=====================================
+
+The `html5lib <https://github.com/html5lib/html5lib-python>`_ module `deprecated <https://github.com/html5lib/html5lib-python/blob/master/CHANGES.rst#11>`_ its own sanitizer in version 1.1. The maintainers "recommend users migrate to Bleach." This tracks the issues encountered in the migration.
+
+Migration path
+==============
+If you upgrade to html5lib 1.1+, you will start getting deprecation warnings when using its sanitizer. If you decide to follow the recommendation and use Bleach for sanitization, you should understand that Bleach has its own sanitizer (despite the fact that it vendors html5lib). You may therefore encounter slightly different results. Avoid Bleach 3.2.0 because it leaks the deprecation warnings (for the sanitizer in the vendored html5lib v1.1); this is fixed in 3.2.1+.
+
+Here is an example of replacing the sanitization method:
+
+.. doctest::
+
+   >>> fragment = "<a href='https://github.com'>good</a> <script>bad();</script>"
+
+   >>> import html5lib
+   >>> html5lib.serialize(html5lib.html5parser.HTMLParser().parseFragment(fragment), sanitize=True)
+   '<a href="https://github.com">good</a> &lt;script&gt;bad();&lt;/script&gt;'
+
+   >>> import bleach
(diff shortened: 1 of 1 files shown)


docs/migrating-from-html5lib-sanitizer.rst
```diff
@@ -0,0 +1,70 @@
...
+
+=====================================
+Migrating from the html5lib sanitizer
+=====================================
+
+The `html5lib <https://github.com/html5lib/html5lib-python>`_ module `deprecated <https://github.com/html5lib/html5lib-python/blob/master/CHANGES.rst#11>`_ its own sanitizer in version 1.1. The maintainers "recommend users migrate to Bleach." This tracks the issues encountered in the migration.
+
+Migration path
+==============
+If you upgrade to html5lib 1.1+, you will start getting deprecation warnings when using its sanitizer. If you decide to follow the recommendation and use Bleach for sanitization, you should understand that Bleach has its own sanitizer (despite the fact that it vendors html5lib). You may therefore encounter slightly different results. Avoid Bleach 3.2.0 because it leaks the deprecation warnings (for the sanitizer in the vendored html5lib v1.1); this is fixed in 3.2.1+.
```
willkg, line 11: This paragraph (and others) should get word wrapped. Among other things, it's easier for me to comment on individual lines in pull requests.

The two sanitizers are very different and have very different goals. The Bleach sanitizer is not a drop-in replacement for the html5lib one. It's almost guaranteed that you will have different results and have to tune the Bleach sanitizer to your needs.

People should always use the latest version of Bleach, so we can remove this last sentence.

```suggestion
If you upgrade to html5lib 1.1+, you may get deprecation warnings when using its sanitizer. If you
follow the recommendation and switch to Bleach for sanitization, you'll need to spend time tuning the
Bleach sanitizer to your needs because the Bleach sanitizer has different goals and is not a drop-in
replacement for the html5lib one.
```

willkg (changes requested):
If you could fix this one paragraph and then make sure the rest of them are word-wrapped, that'd be great.

Thank you for doing this!

willkg:
This looks good! Thank you!

PR #631, opened 2022-01-21
refactoring code with For Else 

zjzh:
refactoring code with For Else which is more pythonic, concise and efficient; how do you think this change which has practical value？


willkg:
This makes a change to a vendored library html5lib. If you want to make this change, you should make it upstream in [](https://github.com/html5lib/html5lib-python).

PR #632, opened 2022-02-04
querystring tests

jvanasco:
Included are three tests, which I believe should pass in the current version of bleach.  They're from an internal suite that is pinned to an earlier release.

The tests just apply `clean` and `linkify` to urls that have (i) a querystring, with (ii) a normal querystring separator (`&`), and (iii) an html encoded separator (`&amp;`).  The expected output based on historical bleach behavior is that BOTH the raw (`&`) and html encoded (`&amp;`) should be encoded to the html encoding (`&amp;`).

Background:

Our application allows users to edit content.  We ran into a double-encoding issue with query args in some contexts.  These tests helped us rule out the issue coming from bleach, as the bleach test-suite did not cover query args.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10-dev", "pypy3"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "pypy3"]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
--- a/tests/test_clean.py
+++ b/tests/test_clean.py
@@ -14,6 +14,18 @@ def test_clean_idempotent():
     assert clean(clean(dirty)) == clean(dirty)
 
 
+def test_clean_idempotent_query():
+    """Make sure that applying the filter twice doesn't change anything."""
+    dirty = '<a href="http://example.com?foo=bar&bar=foo&amp;biz=bash">'
+    assert clean(clean(dirty)) == clean(dirty)
+
+
+def test_clean_idempotent_img():
+    """Make sure that applying the filter twice doesn't change anything."""
+    dirty = '<imr src="http://example.com?foo=bar&bar=foo&amp;biz=bash">'
+    assert clean(clean(dirty, tags=IMG), tags=IMG) == clean(dirty, tags=IMG)
+
+
 def test_only_text_is_cleaned():
     some_text = "text"
     some_type = int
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10-dev", "pypy3"]
```
willkg, line 11: This should be in a different PR that covers 3.10 support.

tests/test_clean.py
```diff
@@ -14,6 +14,18 @@ def test_clean_idempotent():
...
+def test_clean_idempotent_query():
+    """Make sure that applying the filter twice doesn't change anything."""
+    dirty = '<a href="http://example.com?foo=bar&bar=foo&amp;biz=bash">'
+    assert clean(clean(dirty)) == clean(dirty)
+
+
+def test_clean_idempotent_img():
+    """Make sure that applying the filter twice doesn't change anything."""
+    dirty = '<imr src="http://example.com?foo=bar&bar=foo&amp;biz=bash">'
+    assert clean(clean(dirty, tags=IMG), tags=IMG) == clean(dirty, tags=IMG)
```
willkg, line 26: Instead of breaking these out as separate tests, I'd rather we parametrize `test_clean_idempotent`.

tests/test_linkify.py
```diff
@@ -667,6 +667,12 @@ def test_linkify_idempotent():
     assert linkify(linkify(dirty)) == linkify(dirty)
 
 
+def test_linkify_idempotent_query():
+    """Make sure that applying the filter twice doesn't change anything."""
+    dirty = '<a href="http://example.com?foo=bar&bar=foo&amp;biz=bash">'
+    assert linkify(linkify(dirty)) == linkify(dirty)
```
willkg, line 673: I'd rather we parametrize `test_linkify_idempotent`.

.github/workflows/test.yml
willkg: I'm planning to do this soon. Emergencies keep coming up, so I haven't gotten to it, yet.

willkg:
Thank you for doing this!

willkg:
I looked at 3.10 support and it turns out @g-k did it already in 4ce3b35a and you've got an old version of the main branch.

I like the idea behind this PR. If you want, I can close this PR and add the additional tests. That'll reduce the additional work on your part.

Let me know what you want to do.

willkg:
No worries! You do a ton of work on Bleach and I appreciate it. I'll try to get to this this week.

willkg:
I redid this in #641.

PR #642, opened 2022-02-11
Redo of #603 against main / 4.x

jvanasco:
I manually reapplied the patch from the #603 (bleach 3.x) to address #369, as the merge against 4.x did not want to happen cleanly.


willkg (changes requested):
I think this is good except the commit messages aren't very helpful.

Can you squash the commits into a single commit that has a helpful commit message?

willkg:
I grabbed the commits in this PR and redid them in PR #651.

Issue #627, opened 2022-01-05
update goals / non-goals


g-k:
refs #397

- [ ] define supported browsers or at least unsupported browsers (refs: https://github.com/mozilla/bleach/pull/530#issuecomment-1005766395)
  - [ ] document version changes for that set 
  - [ ] consider setting support tiers / levels [similar to what rust does](https://doc.rust-lang.org/nightly/rustc/platform-support.html) e.g. Chrome, Firefox, and Safari release versions are tier 1, nightly version tier 2. Ideally this reflects to E2E testing in CI (similar to DOMPurify) refs #242 
- [ ] update security goals
  - [ ] likely exclude privacy and XS leaks
  - [ ] add threats and goals/non-goals from [DOMPurify](https://github.com/cure53/DOMPurify/wiki/Security-Goals-&-Threat-Model) and [the native sanitizer](https://wicg.github.io/sanitizer-api/#security-considerations) likely with the extensions to reflect that bleach doesn't have access t o the DOM and should sanitize output for multiple browsers. This also puts it in a position to do more about stored and reflected XSS.

Issue #397, opened 2018-10-03
improve bleach documentation


willkg:
I think the documentation feels dense and it lacks good flow. I'm pretty sure a lot of the language could be simplified and reduced to be clearer and more direct.

This issue covers expounding specific issues and then either fixing them or spinning off new issues.

Shivansh2407:
Can I help you on this issue

willkg:
@remote007 What do you mean by MR?

remote007:
Merge Requests . i.e, is it available to work on ?

willkg:
@remote007 I think this issue is lacking definition, so until it's more clearly defined (e.g. what exactly should be done?), it's not a good issue to work on.

willkg:
I'm not going to work on this.

PR #646, opened 2022-02-18
docs: add CSS-based XSS and privacy leak prevention to non-goals

g-k:
fix: #627

This one took me longer to get too than anticipated. 

This PR does not claim support for any browsers in favor of waiting until we can test output in them E2E #242

### DIFF
--- a/docs/goals.rst
+++ b/docs/goals.rst
@@ -150,6 +150,28 @@ that bleach properly strips or escapes language-specific syntax like
 Angular templates before using bleach-sanitized output with your
 framework or template language.
 
+Protect against CSS-based XSS attacks in legacy browsers
+--------------------------------------------------------
+
+Bleach will not protect against CSS-based XSS vectors that only worked
+in legacy IE, Opera, or Netscape/Mozilla/Firefox browsers. For
+example, it will not remove ``expression`` or ``url`` functions in CSS
+component values in style elements or attributes and `other vectors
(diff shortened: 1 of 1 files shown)


docs/goals.rst
```diff
@@ -150,6 +150,28 @@ that bleach properly strips or escapes language-specific syntax like
...
+
+Bleach does not prevent output from fingerprinting users or leaking
+information about users via requests to external sites. For example,
+it will not remove CSS Media Queries or tracking pixels.
+
+See also:
+
+* `browser leaks https://browserleaks.com/`_
+* `HTTP leaks https://github.com/cure53/HTTPLeaks`_
+* `XS leaks https://xsleaks.dev/`_
```
willkg, line 174: We should probably add protecting users from clicking on urls that have malicious or harmful content. I think that's come up a few times and we haven't explicitly stated it's a non-goal.

Also, this non-goals list is pretty long. Maybe there's a better way to think about this? Maybe we can list Bleach's goals and then list everything else in a "These are the other things you should be thinking about that Bleach doesn't cover...." section?

willkg (approved):
Let's land it as it is now.

Before we do a 5.0.0 release, I'll do a pass through the docs (I forget what's in them--it's been a long while) and we can iterate over reducing, reorganizing, and clarifying then.

g-k:
ok sounds good. We have #397 for updating the docs.

Issue #552, opened 2020-09-16
add fuzz / property-based tests


g-k:
e.g. with https://hypothesis.works/

Google-Autofuzz:
I'm currently trying to get bleach integrated with OSS-Fuzz: https://github.com/google/oss-fuzz/pull/4959

@g-k would you be interested in being the primary contact?

g-k:
@Google-Autofuzz yes, I can be the primary contact. Thanks!

Google-Autofuzz:
Can you please give me an email address with an associated google account?

g-k:
How about gguthe@mozilla.com ? Is a gsuite email OK or does it have to be a
gmail account?

PR #647, opened 2022-03-01
Fix character entity handling (#450)

### LINKED ISSUE #450: bleach.clean method returns unexpected value for string with unknown entities

willkg:
This fixes a bug where things in the form of a character entity (starts with &, some letters, and ends with ;) that aren't character entities get mangled. Now they are correctly handled.

This also fixes a bug where things in the form of a character entity that aren't character entities should have had the & escaped, but that wasn't happening.

After this, Bleach does the following:

* valid character entities that start with a & and end with a ; are left alone
* things that start with a & and end with a ; and aren't valid character entities are treated as text and the & is escaped
* things that start with a & but don't end with a ; are treated as text and the & is escaped

This is easy to reason about and idempotent and doesn't suffer from ambiguous edge cases where Bleach can't know what the user intended.

I also cleaned up the comments in the tests so it's clearer what we intend Bleach to be doing.

This fixes #450.

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -512,13 +512,13 @@ def convert_entities(text):
 def match_entity(stream):
     """Returns first entity in stream or None if no entity exists
 
-    Note: For Bleach purposes, entities must start with a "&" and end with
-    a ";". This ignoresambiguous character entities that have no ";" at the
-    end.
+    Note: For Bleach purposes, entities must start with a "&" and end with a
+    ";". This ignores ambiguous character entities that have no ";" at the end.
 
     :arg stream: the character stream
 
-    :returns: ``None`` or the entity string without "&" or ";"
+    :returns: the entity string without "&" or ";" if it's a valid character
+        entity; ``None`` otherwise
 
     """
     # Nix the & at the beginning
@@ -557,13 +557,19 @@ def match_entity(stream):
     # Handle character entities
     while stream and stream[0] not in end_characters:
         c = stream.pop(0)
-        if not ENTITIES_TRIE.has_keys_with_prefix(possible_entity):
-            break
+        print(1, stream, c, possible_entity)
         possible_entity += c
+        if not ENTITIES_TRIE.has_keys_with_prefix(possible_entity):
+            # If it's not a prefix, then it's not an entity and we're
+            # out
+            return None
 
+    print(2, stream, possible_entity)
     if possible_entity and stream and stream[0] == ";":
+        print(3, possible_entity)
         return possible_entity
 
+    print(3, "NONE")
     return None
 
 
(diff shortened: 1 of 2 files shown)


bleach/html5lib_shim.py
```diff
@@ -557,13 +557,19 @@ def match_entity(stream):
...
-        if not ENTITIES_TRIE.has_keys_with_prefix(possible_entity):
-            break
+        print(1, stream, c, possible_entity)
         possible_entity += c
+        if not ENTITIES_TRIE.has_keys_with_prefix(possible_entity):
+            # If it's not a prefix, then it's not an entity and we're
+            # out
+            return None
 
+    print(2, stream, possible_entity)
```
willkg, line 567: Oops--I need to remove the print functions.

bleach/html5lib_shim.py
```diff
@@ -557,9 +557,11 @@ def match_entity(stream):
     # Handle character entities
     while stream and stream[0] not in end_characters:
         c = stream.pop(0)
-        if not ENTITIES_TRIE.has_keys_with_prefix(possible_entity):
-            break
         possible_entity += c
+        if not ENTITIES_TRIE.has_keys_with_prefix(possible_entity):
+            # If it's not a prefix, then it's not an entity and we're
+            # out
+            return None
```
willkg, line 564: This block mirrors the structure of the one that handles numeric entities, but it was awkwardly wrong and had been since I wrote it in 2017:

https://github.com/mozilla/bleach/commit/d8f7d1d651a6b5a4c8e5a5932ea4fd61769a60d4

:(

g-k (approved):
r+

> This is easy to reason about and idempotent and doesn't suffer from ambiguous edge cases where Bleach can't know what the user intended.

Excellent!

Thanks for updating the test comments and writing down the expected behavior. In the future, I'd love to add property-based tests for idempotency and the three behaviors you listed.

willkg:
Oh, that's a good idea. I'll add some more idempotent tests now.

willkg:
@g-k ^^^ Does that cover the idempotent tests you were thinking of? Are there others we should add?

g-k:
> Does that cover the idempotent tests you were thinking of? Are there others we should add?

Oh nice! I was thinking of it as a TODO / good predicates to check for #552

g-k (approved):

willkg:
That makes sense. I haven't quite gotten around to looking at the fuzz stuff, yet. I'll toss that in my list of things to do, but probably after the 5.0.0 release.

Thank you!

Issue #633, opened 2022-02-07
rework how bleach.clean sanitizes css


willkg:
Bleach's `Cleaner` has a `sanitize_css` that's a complete pain in the ass to maintain. Not only is it an inscrutable regex for a complex specification, but it's got lots of edge cases and places where it just plain doesn't work. So far, the way we've been "handling" this is by modifying the regex to fix edge cases.

@g-k mentioned that maybe it's time we took a step back and re-evaluated the goals for sanitizing CSS. What browsers are we targeting? What attacks are we preventing? From there, we can figure out a different course of action. Maybe it's writing or switching to a real css parser. Maybe it's dropping css sanitizing altogether.

g-k:
> What browsers are we targeting?

I'd say either relatively recent Firefox, Safari, and Chromium-based browsers (say back to the last Fx ESR release or two) or not legacy browsers like Trident-based IE and other browsers from that era.

> What attacks are we preventing?

I think this should basically be XSS, DOM Clobbering, and Structural Damage (bleach is Python so XSS via jQuery and Prototype Pollution) i.e. what DOMPurify handles minus the client side specific stuff. Since we expect bleach to run server side mXSS is a larger risk and bleach output should be safe for all supported browsers.

> Maybe it's writing or switching to a real css parser. Maybe it's dropping css sanitizing altogether.

👍 in https://github.com/mozilla/bleach/issues/248 we looked at tinycss2 and some other options. It'll be easier to switch now that we're only on Python 3. Ideally we can provide a "use at your own risk" interface for people to swap in their own logic.

willkg:
If there aren't CSS vectors, then I'd rather not sanitize CSS at all and just take all that code out. CSS only comes into play if someone is using `.clean()` and allowing `style` tags or `style` attributes. Otherwise this isn't an issue with our default safety stuff.

g-k:
My only concern would be that we want to prevent damage to html outside the style attr or style elem.

For the attr, we'd want to validate that it can't be escaped. Doesn't contain quotes or double quotes is probably ok. IIRC we do this for other attrs.

The element is trickier, but maybe we can check that it closes and doesn't contain invalid CSS chars. There might be potential for mXSS if there's a way to use CSS to get html5lib out of sync with how browsers parse the style tag.

willkg:
[...]

Issue #539, opened 2020-06-12
Split API into safe/recipes and hazmat 


g-k:
> I wonder if the ongoing "Bleach does it right by default but then lets
users do whatever with warnings of dire but vague consequences"
architecture changes we're making should also be accompanied by something
that can help users know when their overrides create bad situations.

1. take a cue from https://cryptography.io/en/latest/#layout and split the
API into a safe and potentially a more stable "recipes" layer (just
bleach.clean, bleach.linkify, and maybe some other top-level functions
taking a limited subset of args) and a more dangerous "hazmat" layer with
all the knobs and buttons to tweak and override things. Security bugs
against the recipes layer would be higher severity.

_Originally posted by @g-k in https://github.com/mozilla/bleach/pull/535#issuecomment-642734522_

willkg:
I think this is interesting, but it's probably a fundamental change to Bleach architecture and there are no plans for that, so I'm going to close it out.

PR #648, opened 2022-03-02
WIP: switch to tinycss2 (#633)

willkg:
This switches to using tinycss2 for parsing CSS rules and serializing
CSS tokens after sanitizing.

Things to think about:

1. does this create new problems?
2. does this fix some of the problems (hyphens, etc) we've got in the issue tracker so as to make it worth moving forward with?
3. is the tinycss2 library dependable?
4. should we additionally adjust the `clean` and `Cleaner` APIs to make it easier for users to use their own CSS sanitizers?
5. should we also use this code to sanitize data in `<style>` tags?


bleach/sanitizer.py
```diff
@@ -596,44 +597,34 @@ def disallowed_token(self, token):
...
-|"[\s\w]+"       # a double quoted string of [\s\w]+
-|\([\d,%\.\s]+\) # a parenthesized string of one or more digits, commas, periods, ...
-)*$""",  # ... percent signs, or whitespace e.g. from 'color: hsl(30,100%,50%)'
-            flags=re.U | re.VERBOSE,
-        )
-
-        for part in parts:
-            if not gauntlet.match(part):
-                return ""
+        parsed = tinycss2.parse_declaration_list(style)
```
willkg, line 600: https://doc.courtbouillon.org/tinycss2/stable/api_reference.html#tinycss2.parse_stylesheet_bytes

I wasn't sure whether to enable `trim_whitespace` and `trim_comments`. The output of `trim_whitespace` wipes all the whitespace tokens and it doesn't look good. However, when we remove tokens because they're not in the allow list, we end up with a series of consecutive whitespace tokens. I accounted for that by looking at the last token in the `new_tokens` list (poorly named) to see if it's a whitespace token and ignoring the new one if it is. Maybe that's good enough?

bleach/sanitizer.py
```diff
@@ -596,44 +597,34 @@ def disallowed_token(self, token):
...
+        for token in parsed:
+            if token.type == "at-rule":
+                print("omg")
+            elif token.type == "declaration":
+                if (
+                    token.lower_name in self.allowed_css_properties
+                    or token.lower_name in self.allowed_svg_properties
+                ):
+                    new_tokens.append(token)
+            elif token.type in ("comment", "whitespace"):
```
willkg, line 618: Oh, whoops--this should be split up because we (probably) always want to add the comments.

bleach/sanitizer.py
```diff
@@ -596,44 +597,34 @@ def disallowed_token(self, token):
...
+                if new_tokens and new_tokens[-1].type != token.type:
+                    new_tokens.append(token)
+            # Declaration
+            # AtRule
+            # Comment
+            # WhitespaceToken
+            # ParseError
```
willkg, line 625: This is the list of possible tokens here.

bleach/sanitizer.py
```diff
@@ -596,44 +597,34 @@ def disallowed_token(self, token):
...
-                clean.append(prop + ": " + value + ";")
+        # decl.name.lower() in self.allowed_css_properties
+        # or decl.name.lower() in self.allowed_svg_properties
 
-            elif prop.lower() in self.allowed_svg_properties:
-                clean.append(prop + ": " + value + ";")
+        new_tokens = []
+        for token in parsed:
+            if token.type == "at-rule":
+                print("omg")
```
willkg, line 611: I have no idea what to do with at-rules. For example, `@media`. Would that be an "allowed CSS property"? Should we have "allowed atrules"?

bleach/sanitizer.py
```diff
@@ -596,44 +597,34 @@ def disallowed_token(self, token):
...
+            # AtRule
+            # Comment
+            # WhitespaceToken
+            # ParseError
+
+        if not new_tokens:
+            return ""
 
-        return " ".join(clean)
+        return tinycss2.serialize(new_tokens).strip()
```
willkg, line 630: I strip whitespace at the beginning and ending because it cleans things up and I couldn't think of a situation where this would be unhelpful.

Are there situations where this is unhelpful?

setup.py
```diff
@@ -24,6 +24,7 @@ def get_version():
 
 
 INSTALL_REQUIRES = [
+    "tinycss2>=1.1.0",
```
willkg, line 27: I need to add an upper bound here.

tests/test_css.py
```diff
@@ -110,40 +110,39 @@ def test_valid_css():
...
         ),
         (
             "<p style=\"background: url(  'topbanner.png'  ) #00D;\">foo</p>",
-            '<p style="background: #00D;">foo</p>',
+            "<p style='background: url(  \"topbanner.png\"  ) #00D;'>foo</p>",
         ),
         (
             "<p style=\"background: url (  'topbanner.png'  ) #00D;\">foo</p>",
-            '<p style="background: #00D;">foo</p>',
+            "<p style='background: url (  \"topbanner.png\"  ) #00D;'>foo</p>",
```
willkg, line 140: I'm not sure we need all these variations anymore, but I didn't spend the time to think through what they're actually testing with the new code.

tests/test_css.py
```diff
@@ -110,40 +110,39 @@ def test_valid_css():
...
         ),
         # Verify urls with character entities
         (
             "<p style=\"background: url&#x09;('topbanner.png') #00D;\">foo</p>",
-            '<p style="background: #00D;">foo</p>',
+            '<p style="background: url&#x09;">foo</p>',
```
willkg, line 145: This change is weird, but I'm not sure what we should expect here. Maybe the test itself is unhelpful?

tests/test_css.py
```diff
@@ -239,9 +238,11 @@ def test_style_hang():
...
         ),
         (
-            '<p style="font-family: &quot;Droid Sans&quot;, serif; white-space: pre-wrap;">text</p>',
-            ["font-family", "white-space"],
-            "<p style='font-family: \"Droid Sans\", serif; white-space: pre-wrap;'>text</p>",
+            "<p style=\"content: '&nbsp;';\">text</p>",
+            ["content"],
+            # This seems wrong, but it's right. Character entities aren't a
+            # thing in CSS, so this is treated as text.
+            "<p style='content: \"&amp;nbsp;\";'>text</p>",
```
willkg, line 245: I think this is right, but it'd be nice to have a second set of eyes.

willkg:
@jvanasco @stucox @FelixSchwarz @kiilerix @Alex3917 @tony @EmilStenstrom @t-ionut and anyone else following along with sanitize css issues in Bleach, does this work for you? What else should we be thinking about and testing?

Alex3917:
Thanks @willkg, this looks great to me! It seems to be ~5% faster also. The only difference I notice over the ~400 tests I have for this is just the spaces being collapsed after colons, so e.g. 'font-space: 700' becoming 'font-space:700'. But that's fine with me!

kiilerix:
> ... @kiilerix ...  does this work for you? What else should we be thinking about and testing?

What is the goal of the CSS sanitization? I don't see anything in goals.rst that justifies it. If we trust browsers to handle garbage "safely" and don't care about "looks", what is left? And without an explicit statement of the goal, it is hard to comment on whether this change achieves it or not.

The problems I saw are fixed already by removing the previous unnecessarily strict and arbitrary CSS validation / sanitization.

I'm not sure about the use of tinycss2. While it presumably is "safe" for parsing arbitrary CSS, it might not provide the expected safety properties when used in this way. It is extra complexity and dependency. Does it really add any kind of actual security that justifies it?

I think I'm +1 for dropping the old code and -0 for introducing tinycss2.

FelixSchwarz:
I used tinycss2 in the past without problems (though I did not really do complicated stuff using the library). From a Fedora perspective we have v1.1+ in all supported Fedora releases and tinycss2 1.0.x in EPEL 8.

jvanasco:
Our main application is a Content Management System.  On a given "property" there are two general types of users: "admin" and "public".

We completely block inline CSS with bleach for security and UX concerns for all users in all contexts. I could not imagine a way to allow it in user-generated content that did not introduce issues.

We allow "admins" to upload custom templates. Those are cleaned by bleach and allow `class` attributes for CSS customization, and only injected into the public facing content templates – for security concerns there is a strict separation in where those can be injected, so it can not affect anything touching account or property management.  Those are parsed by another library, I think it is tinycss.  We created an overly complex way of implementing and restricting the CSS overrides with the template structure and load order, I don't recall the specifics offhand.

@Alex3917 Additional concerns with CSS include breaking UX, disrupting comment systems, and sneaking in tracking pixels or images via "url". The latter the same vector as described for data exfiltration in the link you shared, but something I saw attempted often by bots and scammers when I ran a large website with a highly active comment system.

g-k (commented):
Ah this took me longer than expected to get to, but thanks @willkg for pushing this up and everyone for their feedback.

> 1. does this create new problems?

It is a new dependency for non-core functionality, since we're not worrying about CSS-based XSS for legacy browsers in 5.x.

It could make UI redress and privacy or data leaks easier. Bleach doesn't have visibility into how its output is displayed so it can't fix UI redress (I should document this). Bleach drops style attrs and tags by default, so advanced users customizing their bleach config should be aware of the risks associated with CSS they allow (better docs and #539 could help here).

Fixing CSS related bugs, reducing regex-related maintenance in the future, and the perf improvements Alex saw would make it a net win.

> 2. does this fix some of the problems (hyphens, etc) we've got in the issue tracker so as to make it worth moving forward with?

We should test those out. I'll try to tack on some xfail test cases for those issues and hopefully we can get them passing here.

> 3. is the tinycss2 library dependable?

I haven't used it extensively, but it seems like a reasonable option. Shipping in Fedora is a positive indicator.

> 4. should we additionally adjust the `clean` and `Cleaner` APIs to make it easier for users to use their own CSS sanitizers?

Ideally yes, but it doesn't have to make it into 5.x.

> 5. should we also use this code to sanitize data in `<style>` tags?

👍  they'd both use parse declaration list

> 6. should we make tinycss2 an extras install? then it's not required for users who aren't sanitizing css.

Yes, it'd be nice to reduce download size, improve modularity, and split out non-core functionality. Assuming tinycss2 stays relatively stable we should be able to avoid vendoring it like we had to for html5lib.

willkg:
^^^ That improves things a bit.

1. It makes css sanitization require the `css` extras which will install tinycss2. Thus tinycss2 is now an optional dependency. It specifies an upper and lower bound for acceptable versions.
2. It redoes the API around css sanitization so it's done with a class. This makes it easier for devs to drop in a different class if they need different functionality.
3. It defaults to ALLOWED_CSS_PROPERTIES that html5lib sanitizer used rather than the empty list.
4. It updates the documentation and tests accordingly.

I spent some time looking at adding support for sanitizing `<style>` content. I think I'm going to push that off.

I still need to figure out AtRule handling.

willkg:
I'm going to punt on the AtRule handling. It seems nice for completeness, but I don't have any good ideas on how to do it and if someone needs it, they can write their own CSSSanitizer since they probably have other needs as well. Given that, I think I'm done working on this and it's ready for review.

g-k (approved):
r+ with the nit that we need to drop the omg print in css parsing.

It's great to see the CSS parsing regexes go and associated regressions fixed.

+1 to using the html5lib sanitizer `ALLOWED_CSS_PROPERTIES` that should make https://github.com/html5lib/html5lib-python/issues/443 easier for people.

docs/clean.rst
```diff
@@ -5,25 +5,20 @@
...
+algorithm and sanitize tags, attributes, and other aspects. This also handles
+unescaped characters and unclosed and misnested tags. The result is text that
+can be used in HTML as is.
 
 .. Warning::
 
    :py:func:`bleach.clean` is for sanitising HTML fragments to use in an HTML
-   context--not for HTML attributes, CSS, JSON, xhtml, SVG, or other contexts.
+   context--not for use in HTML attributes, CSS, JSON, xhtml, SVG, or other
+   contexts.
```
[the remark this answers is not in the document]
willkg: Yes. We probably should. I'll fix that.

willkg:
@g-k Thank you for helping me through that!

bleach/css_sanitizer.py
```diff
@@ -0,0 +1,108 @@
...
+        """Sanitizes css in style tags"""
+        parsed = tinycss2.parse_declaration_list(style)
+
+        if not parsed:
+            return ""
+
+        new_tokens = []
+        for token in parsed:
+            if token.type == "at-rule":
+                print("omg")
```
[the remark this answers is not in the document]
EmilStenstrom: Would commenting it out, but keeping it in the content, create risk?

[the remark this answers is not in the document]
g-k: I don't think it'd introduce much risk, since it still needs to be a valid at-rule.

If there's demand, maybe we should introduce hooks by token type to make it easier to derive tinycss2 parsers without copying of most or all of `css_sanitizer.py`.

bleach/sanitizer.py
g-k: @stucox interesting! Can you share what the media queries targeted (e.g. devices, different authors) and what kind of config you used for them or would've liked to use?

PR #651, opened 2022-04-04
fix stripping block-level tags (#369)

### LINKED ISSUE #369: [Accessibility bug] Semantic whitespace implied by block elements isn't retained properly

willkg:
When Bleach strips a block-level element, it should replace it with a newline preserving the whitespace that would exist if it was being parsed by a browser.

Fixes #369.

This supersedes PR #642.


tests/test_clean.py
```diff
@@ -1070,6 +1070,42 @@ def test_html_comments_escaped(namespace_tag, end_tag, eject_tag, data, expected
...
+            "<p>Te<b>st</b>!</p><p>Hello</p>",
+            "Test!\nHello",
+        ),
+        (
+            # with an internal space and escaped character
+            "<p>This is our <b>description!</b> &amp;</p><p>nice!</p>",
+            "This is our description! &amp;\nnice!",
+        ),
+        (
+            # double-wrap causes an initial newline--this can"t really be
```
g-k, line 1086: extra nitty nit: the apostrophe is rendering with a double quote in `can"t`

g-k (approved):
r+ lgtm

Alex3917:
Does it make sense to make this optional? (Other than via monkey patching `HTML_TAGS_BLOCK_LEVEL`.)

willkg:
> Does it make sense to make this optional? (Other than via monkey patching `HTML_TAGS_BLOCK_LEVEL`.)

I'm going to evaluate "does it make sense" in terms of "users have expressed a need" and so far I haven't seen anyone express a need to make this optional. In my uses of Bleach, having the additional newlines would have been fine. Given that, I think the answer at this time should be no. If someone has a need, they can write up an issue and we can evaluate things from there.

PR #659, opened 2022-05-03
Correctly urlencode email address parts

### LINKED ISSUE #658: bug: `linkify` with `parse_email=True` doesn't handle "%" a "?" in `addr-specs`
**Describe the bug**

bug: `linkify` with `parse_email=True` doesn't handle "%" and "?", which may occur in RFC822 addr-specs (see https://datatracker.ietf.org/doc/html/rfc2368#section-6)

 - Python Version: 3.10.4
 - Bleach Version: 5.0.0

**To Reproduce**

```python
>>> bleach.linkify("gorby%kremvax@example.com", parse_email=True)
'<a href="mailto:gorby%kremvax@example.com">gorby%kremvax@example.com</a>'
```

**Expected behavior**

I expected RFC822 special characters to be percent-encoded according to RFC2368:

```python
>>> bleach.linkify("gorby%kremvax@example.com", parse_email=True)
'<a href="mailto:gorby%25kremvax@example.com">gorby%kremvax@example.com</a>'
```

**Additional context**

Same issue exists with "?"; I didn't test other RFC822 special characters but suspect they are similarly left unquoted.

larseggert:
Fixes #658

### DIFF
--- a/bleach/linkifier.py
+++ b/bleach/linkifier.py
@@ -308,7 +308,7 @@ class LinkifyFilter(html5lib_shim.Filter):
                     # Run attributes through the callbacks to see what we
                     # should do with this match
                     attrs = {
-                        (None, "href"): "mailto:%s" % address),
+                        (None, "href"): "mailto:%s" % address,
                         "_text": match.group(0),
                     }
                     attrs = self.apply_callbacks(attrs, True)
--- a/bleach/linkifier.py
+++ b/bleach/linkifier.py
@@ -300,10 +300,15 @@ class LinkifyFilter(html5lib_shim.Filter):
                             {"type": "Characters", "data": text[end : match.start()]}
                         )
 
+                    # URL-encode the "local-part" according to RFC6068
+                    parts = match.group(0).split("@")
+                    parts[0] = quote(parts[0])
+                    address = "@".join(parts)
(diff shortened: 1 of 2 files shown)


willkg:
@larseggert Can you fix up the broken tests? Thanks!

larseggert:
@willkg could you kick the CI tests? For some reason, running the tests locally via `tox` doesn't work.

willkg:
You should be able to run tests locally. Can you elaborate on what doesn't work when you run tox?

willkg:
If you create a virtual environment and install everything with:

```
pip install -e '.[dev,css]'
```

that should install Bleach into the virtual environment in an editable mode which will reflect changes you're making in the `src/` directory.

Does that help?

willkg:
Python comes with a virtual environment system. It's a convenient way to install things in a sandbox that doesn't affect your system-wide Python setup. This is very convenient for development where the things you're working on might be in flux or have bugs.

* https://docs.python.org/3/library/venv.html
* https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment
* https://bleach.readthedocs.io/en/latest/dev.html#install-for-development

Issue #668, opened 2022-06-02
task: check bleach url sanitizing and urlparse against wpt url tests


willkg:
From @g-k comment in issue #662:

> Checking bleach and python urlparse behavior against the [wpt url tests](https://github.com/whatwg/url#tests-and-implementations) would probably be good too.

willkg:
This is the file I was looking through: https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json

willkg:
I'm not going to do this. Closing it out.

PR #671, opened 2022-06-03
add urlparse tests based on wpt url tests (#668)

willkg:
This adds 20 of the wpt url tests. There are like a thousand of them. We probably should add more, but it needs to be done manually because the parse results have a different shape and theirs factor in a base url. I added comments on some output differences I saw.

### DIFF
--- a/tests/test_parse_shim.py
+++ b/tests/test_parse_shim.py
@@ -1,0 +1,101 @@
+from dataclasses import dataclass
+
+import pytest
+
+from bleach.parse_shim import urlparse
+
+
+@dataclass
+class ParseResult:
+    scheme: str = ""
+    netloc: str = ""
+    path: str = ""
+    params: str = ""
+    query: str = ""
+    fragment: str = ""
+
+
+# Tests from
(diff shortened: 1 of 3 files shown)


willkg:
@g-k Is this along the lines of what you were thinking? If you skim through the file, what other ones would be good to add in this pass?

tests/test_parse_shim.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+@dataclass
+class ParseResult:
+    scheme: str = ""
+    netloc: str = ""
+    path: str = ""
+    params: str = ""
+    query: str = ""
+    fragment: str = ""
```
g-k, line 15: Nice! I think this makes the table-based tests cleaner.

Possible future TODO would be to capture errors in this class like a Rust Result, but I guess Python usually avoids that?

g-k (approved):
r+ lgtm thanks for tackling this!

tests/test_parse_shim.py
willkg: Oh, that's a good idea. I'll add a failure test and see how that might work.

PR #683, opened 2022-11-25
Update tinycss2 1.2.0

olemoign:
The only breaking change is dropping support of Python 3.6, which bleach already dropped.


taylor-cedar:
Would it be possible to merge this?

OttoAndrey (approved):

PR #687, opened 2022-12-08
Switch to requirements-dev.txt for dev requirements

willkg:
This also removes "python setup.py build" from the tinycss2 tox environments--we don't need to test that in those environments.


tox.ini
```diff
@@ -11,46 +11,46 @@ envlist =
...
     pytest {posargs:-v}
     python setup.py build
 
 [testenv:py37-tinycss2]
-extras = dev,css
+deps = -rrequirements-dev.txt
+extras = css
 commands =
     pytest {posargs:-v}
-    python setup.py build
```
willkg, line 23: I don't think these tox environments need to run `python setup.py build` since it's already run in the regular non-tinycss2 environment and adding tinycss2 to the mix doesn't affect builds at all. So in the interest of reducing test time, I nixed this.

requirements-dev.txt
```diff
@@ -0,0 +1,12 @@
...
+black==22.3.0; implementation_name == 'cpython'
+build==0.8.0
+flake8==4.0.1
+mypy==0.961; implementation_name =='cpython'
+pip-tools==6.6.2
+pytest==7.1.2
+Sphinx==4.3.2
+tox==3.25.0
+twine==4.0.1
+wheel==0.37.1
```
willkg, line 12: I'm going to update these next.

PR #691, opened 2022-12-23
Add warning when css_sanitizer is not set, but style is allowed (#676)

### LINKED ISSUE #676: bug: bleach truncates Katex style attributes
**Bleach truncates a lot of Katex style attributes**

Basic example: a markdown_katex output may contain a span like so : `<span class="vlist" style="height:1.0697em;">`. It contains a `style` attribute, and when passed through bleach (allowing the `style` attribute), I get this:

```
<span class="vlist" style=""></span>
```

While the desired output would be:

```
<span class="vlist" style="height:1.0697em;"></span>
```

As a result, actual Katex math doesn't render properly.

** python and bleach versions:**

 - Python Version: 3.9.2
 - Bleach Version: 5.0.1
 - Markdown-katex Version: 202112.1034

**To Reproduce**

```python
from bleach import Cleaner
 
cleaner = Cleaner(tags = ['span'],
                  attributes = {'span': ['class', 'style']})
          
minimal_katex_span = '<span class="vlist" style="height:1.0697em;">'
res = cleaner.clean(minimal_katex_span)
print(res)
```

**Additional context**

I am unsure if this is actually a bug or intended behavior in some way. The more general problem I face is: how to correctly use bleach after user input is transformed through markdown with the markdown_katex extension?

willkg:
This adds a note to the documentation and also adds a warning that gets emitted if css_sanitizer is not set, but "style" is an allowed attribute. This reduces the likelihood of developer error.

Fixes #676.

### DIFF
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -48,6 +48,10 @@ INVISIBLE_CHARACTERS_RE = re.compile("[" + INVISIBLE_CHARACTERS + "]", re.UNICOD
 INVISIBLE_REPLACEMENT_CHAR = "?"
 
 
+class NoCssSanitizerWarning(UserWarning):
+    pass
+
+
 class Cleaner:
     """Cleaner for cleaning HTML fragments of malicious content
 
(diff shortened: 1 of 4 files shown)


bleach/sanitizer.py
```diff
@@ -143,6 +147,25 @@ def __init__(
...
+                attributes_values = []
+                for values in attributes.values():
+                    if isinstance(values, (list, tuple)):
+                        attributes_values.extend(values)
+
+            if "style" in attributes_values:
+                warnings.warn(
+                    "'style' attribute specified, but css_sanitizer not set.",
+                    category=NoCssSanitizerWarning,
+                )
```
willkg, line 167: Mmm... I should probably do this in the sanitizer. I'll move it.

willkg: On second thought, keeping it here means that it only runs once per cleaner and not once per clean, so I'll keep it here.

PR #693, opened 2023-01-13
Change elements to tags (#649)

### LINKED ISSUE #649: bug: allowed_elements in BleachSanitizerFilter is different from html5lib_shim.SanitizerFilter
**Describe the bug**

I was trying to replace `html5lib.filters.sanitizer.Filter` with `bleach.sanitizer.BleachSanitizerFilter` (which inherits from the earlier one). We used that filter by just doing `Filter(stream)`. And we find that we're getting all the tags alphabetized, which we don't really want.

I found that to know if something needs to be alphabetized, there's this check:
https://github.com/mozilla/bleach/blob/e4718bdda1c43849a8dfbda1b2915e8a6594aab2/bleach/sanitizer.py#L354

That is only set in the `html5lib.filters.sanitizer.Filter` `__init__` function by sending an `allowed_elements`. It also has some default ones. But the default value for that is a set of tuples `(namespace, token name)`, not some iterable of token names:
https://github.com/html5lib/html5lib-python/blob/f87487a4ada2d6cf223bdd182774a01ba3c84618/html5lib/filters/sanitizer.py#L31-L32

And that is used like this:
https://github.com/html5lib/html5lib-python/blob/f87487a4ada2d6cf223bdd182774a01ba3c84618/html5lib/filters/sanitizer.py#L808
```py
            if ((namespace, name) in self.allowed_elements or
                (namespace is None and
                 (namespaces["html"], name) in self.allowed_elements)):
```

I see that `allowed_elements` is overriden in the `Cleaner.clean(text)` method, and a list of tags is provided.
https://github.com/mozilla/bleach/blob/e4718bdda1c43849a8dfbda1b2915e8a6594aab2/bleach/sanitizer.py#L179

So that is consistent with the usage of the `BleachSanitizerFilter` `sanitize_token` method, but that is not a consistent format with the parent.

**python and bleach versions (please complete the following information):**

 - Python Version: 3.7
 - Bleach Version: 2.3.1

**Expected behavior**

I propose one of these 3 options:
- `BleachSanitizerFilter` should use `allowed_elements` that are compatible with `html5lib` ones, which would require a lot of refactoring and it's likely to not be backwards compatible
- `BleachSanitizerFilter` should force clients to send a new compatible set of `allowed_elements` by adding it as a required argument, as it's not really optional to not send it because it's not compatible with the usage
   ```py
   class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
       """html5lib Filter that sanitizes text

       This filter can be used anywhere html5lib filters can be used.

       """
       def __init__(
           self,
           source,
           allowed_elements,
           attributes=ALLOWED_ATTRIBUTES,
           strip_disallowed_elements=False,
           strip_html_comments=True,
           **kwargs,
       ):
   ```
- `BleachSanitizerFilter` should accept both formats of `allowed_elements` by doing something like this:
   ```py
   def _is_token_allowed(token):
       name = token["name"]
       namespace = token.get("namespace")
       first_allowed_element = next(iter(self.allowed_elements))

       if isinstance(first_allowed_element, tuple):
           return (
               (namespace, name) in self.allowed_elements
               or (
                   namespace is None and
                   and (namespaces["html"], name) in self.allowed_elements
               )
           )
       return name in self.allowed_elements
   ```

willkg:
This cleans up some API weirdness where `BleachSanitizerFilter` subclasses html5lib's sanitizer filter and sets `allowed_elements` to something entirely different than html5lib's sanitizer filter's `allowed_elements` is. The formers is a list of strings. The latter is a frozenset of `(namespace, element)` tuples.

This fixes that by changing the `BleachSanitizerFilter` code to use `allowed_tags` everywhere. This eliminates the API confusion.

While doing this, I changed `ALLOWED_TAGS` to be a frozenset, changed allowed_tags to take a set of strings rather than a list of strings, and renamed `strip_disallowed_elements` to `strip_disallowed_tags` to be consistent with naming. This is a backwards incompatible change we'll need to mention in `CHANGES` and it means a major version change.

I also changed the `__init__` for `BleachSanitizerFilter` so it no longer calls the html5lib sanitizer filter `__init__` which kicks up the deprecation warning.

Fixes #649.

### DIFF
--- a/bleach/html5lib_shim.py
+++ b/bleach/html5lib_shim.py
@@ -38,6 +38,9 @@ from bleach._vendor.html5lib.filters.sanitizer import (
     allowed_protocols,
     allowed_css_properties,
     allowed_svg_properties,
+    attr_val_is_uri,
+    svg_attr_val_allows_ref,
+    svg_allow_local_href,
 )  # noqa: E402 module level import not at top of file
 from bleach._vendor.html5lib.filters.sanitizer import (
     Filter as SanitizerFilter,
--- a/bleach/sanitizer.py
+++ b/bleach/sanitizer.py
@@ -8,21 +8,23 @@ from bleach import html5lib_shim
 from bleach import parse_shim
 
 
-#: List of allowed tags
-ALLOWED_TAGS = [
-    "a",
-    "abbr",
-    "acronym",
-    "b",
-    "blockquote",
-    "code",
-    "em",
-    "i",
-    "li",
-    "ol",
-    "strong",
-    "ul",
-]
+#: Set of allowed tags
+ALLOWED_TAGS = frozenset(
+    (
+        "a",
+        "abbr",
+        "acronym",
+        "b",
+        "blockquote",
+        "code",
+        "em",
+        "i",
+        "li",
+        "ol",
+        "strong",
+        "ul",
+    )
+)
 
 
 #: Map of allowed attributes by tag
@@ -93,7 +95,7 @@ class Cleaner:
     ):
         """Initializes a Cleaner
 
-        :arg list tags: allowed list of tags; defaults to
+        :arg set tags: set of allowed tags; defaults to
             ``bleach.sanitizer.ALLOWED_TAGS``
 
         :arg dict attributes: allowed attributes; can be a callable, list or dict;
@@ -190,13 +192,11 @@ class Cleaner:
         dom = self.parser.parseFragment(text)
         filtered = BleachSanitizerFilter(
             source=self.walker(dom),
-            # Bleach-sanitizer-specific things
+            allowed_tags=self.tags,
             attributes=self.attributes,
-            strip_disallowed_elements=self.strip,
+            strip_disallowed_tags=self.strip,
             strip_html_comments=self.strip_comments,
             css_sanitizer=self.css_sanitizer,
-            # html5lib-sanitizer things
-            allowed_elements=self.tags,
             allowed_protocols=self.protocols,
         )
 
@@ -260,19 +260,21 @@ class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
     def __init__(
         self,
         source,
-        allowed_elements=ALLOWED_TAGS,
+        allowed_tags=ALLOWED_TAGS,
         attributes=ALLOWED_ATTRIBUTES,
         allowed_protocols=ALLOWED_PROTOCOLS,
-        strip_disallowed_elements=False,
+        attr_val_is_uri=html5lib_shim.attr_val_is_uri,
+        svg_attr_val_allows_ref=html5lib_shim.svg_attr_val_allows_ref,
+        svg_allow_local_href=html5lib_shim.svg_allow_local_href,
+        strip_disallowed_tags=False,
         strip_html_comments=True,
         css_sanitizer=None,
-        **kwargs,
     ):
         """Creates a BleachSanitizerFilter instance
 
         :arg source: html5lib TreeWalker stream as an html5lib TreeWalker
 
-        :arg list allowed_elements: allowed list of tags; defaults to
+        :arg set allowed_tags: set of allowed tags; defaults to
             ``bleach.sanitizer.ALLOWED_TAGS``
 
         :arg dict attributes: allowed attributes; can be a callable, list or dict;
@@ -281,8 +283,16 @@ class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
         :arg list allowed_protocols: allowed list of protocols for links; defaults
             to ``bleach.sanitizer.ALLOWED_PROTOCOLS``
 
-        :arg bool strip_disallowed_elements: whether or not to strip disallowed
-            elements
+        :arg attr_val_is_uri: set of attributes that have URI values
+
+        :arg svg_attr_val_allows_ref: set of SVG attributes that can have
+            references
+
+        :arg svg_allow_local_href: set of SVG elements that can have local
+            hrefs
+
+        :arg bool strip_disallowed_tags: whether or not to strip disallowed
+            tags
 
         :arg bool strip_html_comments: whether or not to strip HTML comments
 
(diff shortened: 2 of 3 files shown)


bleach/sanitizer.py
```diff
@@ -8,21 +8,23 @@
...
-    "code",
-    "em",
-    "i",
-    "li",
-    "ol",
-    "strong",
-    "ul",
-]
+#: Set of allowed tags
+ALLOWED_TAGS = frozenset(
```
willkg, line 12: I thought about going through and making the other constant values frozensets where applicable, too. Seems like a good idea. Thoughts?

willkg:
@g-k I think the API weirdness in issue #649 is worth fixing. This was the minimally viable way I could think of fixing it that didn't create a slew of backwards-incompatible changes.

Other options:

1. Change `allowed_tags` to `allowed_elements` (and generally uses of the word "tags" to "elements", too) everywhere and change it to be a set of `(namespace, element)` tuples. This changes the argument name everywhere, but also, developers would have to add namespaces to their `allowed_elements` values. Further, this interacts with `allowed_attributes` in that the keys now need to have elements and callable values would need to support namespaces, too. I remember html5lib adding namespace support, but that it wasn't complete in some way, so this might not even work right. It's a lot of changes for us and a lot of changes for users.
2. Pull in what little bits of code we're still using from html5lib sanitizer filter into our `BleachSanitizerFilter` so we're not longer subclassing it and then clean up the API. This is a lot of work for us plus we have to deal with how to pull in the code without violating license/copyright in a way that's maintainable.
3. Adjust the constructors so that if `allowed_tags` is a list/set of `(namespace, element)` tuples, drop the namespaces. This feels like it creates a potential surprise for users and would warrant someone writing up another bug.
4. Do nothing. I think this is worth fixing, so we could do nothing, but I'll feel bad for the rest of my life.

What do you think? Does this look ok?

bleach/sanitizer.py
```diff
@@ -260,19 +260,21 @@ class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
...
         attributes=ALLOWED_ATTRIBUTES,
         allowed_protocols=ALLOWED_PROTOCOLS,
-        strip_disallowed_elements=False,
+        attr_val_is_uri=html5lib_shim.attr_val_is_uri,
+        svg_attr_val_allows_ref=html5lib_shim.svg_attr_val_allows_ref,
+        svg_allow_local_href=html5lib_shim.svg_allow_local_href,
+        strip_disallowed_tags=False,
         strip_html_comments=True,
         css_sanitizer=None,
-        **kwargs,
```
willkg, line 269: Getting rid of this undocumented pass-through is a good idea and makes this easier to reason about.

bleach/html5lib_shim.py
```diff
@@ -38,6 +38,9 @@
     allowed_protocols,
     allowed_css_properties,
     allowed_svg_properties,
+    attr_val_is_uri,
+    svg_attr_val_allows_ref,
+    svg_allow_local_href,
```
g-k, line 43: ah I see these get re-imported in sanitizer

bleach/sanitizer.py
g-k: 👍 sgtm. I think we used lists for backwards compat or as a holdover from Python 2 support.

bleach/sanitizer.py
```diff
@@ -190,13 +192,11 @@ def clean(self, text):
         dom = self.parser.parseFragment(text)
         filtered = BleachSanitizerFilter(
             source=self.walker(dom),
-            # Bleach-sanitizer-specific things
```
g-k, line 193: 🎉

bleach/sanitizer.py
```diff
@@ -260,19 +260,21 @@ class BleachSanitizerFilter(html5lib_shim.SanitizerFilter):
     def __init__(
         self,
         source,
-        allowed_elements=ALLOWED_TAGS,
```
g-k, line 263: good to fix this rename

bleach/sanitizer.py
```diff
@@ -281,33 +283,41 @@ def __init__(
...
 
         :arg CSSSanitizer css_sanitizer: instance with a "sanitize_css" method for
             sanitizing style attribute values and style text; defaults to None
 
         """
+        # NOTE(willkg): This is the superclass of
+        # html5lib.filters.sanitizer.Filter. We call this directly skipping the
+        # __init__ for html5lib.filters.sanitizer.Filter because that does
+        # things we don't need to do and kicks up the deprecation warning for
+        # using Sanitizer.
```
g-k, line 307: 👍

bleach/sanitizer.py
```diff
@@ -281,33 +283,41 @@ def __init__(
...
-            message="html5lib's sanitizer is deprecated",
-            category=DeprecationWarning,
-            module="bleach._vendor.html5lib",
-        )
-        return super().__init__(
-            source,
-            allowed_elements=allowed_elements,
-            allowed_protocols=allowed_protocols,
-            **kwargs,
-        )
```
g-k, line 310: 💯 nice to drop this hack

g-k (approved):
r+

bleach/sanitizer.py
willkg: I'll look at doing that in another PR.

willkg:
Thank you!

bleach/sanitizer.py
g-k: I think the set is a better data structure and we're bumping the major version. I was just trying to recall context on why it was a list.

bleach/sanitizer.py
NeilGirdhar: Can't help but comment on this, but another way to accomplish what you want while still calling super would be to have the `html5lib.filters.sanitizer.Filter.__init__` call a private method `_f` that does some of initialization.  Then, here in the derived class, you can override `_f` to not do that stuff.

willkg: html5lib is vendored, so we shouldn't modify it.

PR #730, opened 2024-03-22
really drop python<3.7 support

kloczek:
Already in pyproject.toml is set support for python>=3.8 however still is used oler syntax.
Filet all python code over `pyupgrade --py3.8-plus`.


willkg:
We can't change vendored files. If you want to change vendored files (everything in `bleach/_vendor/`), please do it upstream.

kloczek:
Hmm .. but why at all `html5lib` is vendored? 🤔

PR #735, opened 2024-08-02
Add support for some additonal TLDs at linkifier.py

NannoSilver:
Add support for some additional top level domains picked from [https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains](https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains)


willkg:
This project is deprecated. You can pass in a tlds list to `build_url_re` to solve your immediate need. We're not accepting changes like this anymore. Sorry!

PR #742, opened 2024-10-28
drop six dependency (#618)

### LINKED ISSUE #618: remove six from dependencies
Since support for Python 2 was removed in [version 2.0](https://github.com/mozilla/bleach/blob/a362171055af36996dc1cedb64c6b3f68359db27/CHANGES#L548) the use of SIX library to handle compatibility between Python 3 and Python 2 has no more meaning (aka: it is a useless dependecy).

See [remaining code using SIX](https://github.com/mozilla/bleach/search?q=six).

willkg:
This does some squirrely things in order to drop the six dependency that our vendored html5lib has. The thinking is along these lines:

1. this is the best way to drop the six dependency in the context we have
2. it maintains the rigor we use for vendored libraries
3. it's mostly straightforward and easy to reason about
4. the code changes are minimal--it mostly teaks a small handful of import lines
5. it won't cause problems with other things that are installed alongside bleach and its vendored html5lib
6. html5lib hasn't had a release in a while so this won't meaningfully increase the maintenance burden

The way this works is that vendoring now has two steps:

1. installing html5lib 1.1 into a specific directory with no dependencies
2. applying the `01_html5lib_six.patch`

Fixes #618.

### DIFF
--- a/bleach/six_shim.py
+++ b/bleach/six_shim.py
@@ -1,0 +1,19 @@
+"""
+Replacement module for what html5lib uses six for.
+"""
+
+import http.client
+import operator
+import urllib
+
+
+PY3 = True
+binary_type = bytes
+string_types = (str,)
+text_type = str
+unichr = chr
+viewkeys = operator.methodcaller("keys")
+
+http_client = http.client
+urllib = urllib
+urllib_parse = urllib.parse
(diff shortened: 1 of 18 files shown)


bleach/six_shim.py
```diff
@@ -0,0 +1,49 @@
...
+text_type = str
+unichr = chr
+viewkeys = operator.methodcaller("keys")
+
+http_client = http.client
+urllib = urllib
+urllib_parse = urllib.parse
+
+
+# Copied from six: https://github.com/benjaminp/six/blob/c1b416f24de52ebd8eaed3e379ab85ec9c3e935b/six.py
```
willkg, line 24: I don't need this. I'll fix it.

sins921 (approved):

PR #744, opened 2025-02-03
Update Doc Clean Method

goharattiq:


bleach/__init__.py
```diff
@@ -58,7 +58,7 @@ def clean(
     :arg dict attributes: allowed attributes; can be a callable, list or dict;
         defaults to ``bleach.sanitizer.ALLOWED_ATTRIBUTES``
 
-    :arg list protocols: allowed list of protocols for links; defaults
+    :arg set protocols: set of allowed protocols for links; defaults
```
willkg, line 61: Can you update the other instances where this is wrong?

goharattiq: @willkg I have updated the document for `protocols` where required. Let's have a look, please.

willkg (approved):
Thank you for doing this!

PR #783, opened 2026-04-28
fix: fix tox platform restrictions

willkg:
Several environments were restricted to running only on Linux. I'm not really sure why and it prevents me from running tests on macOS. I removed the restriction to see if it passes CI. If it does, we'll keep this fix.


willkg:
vendorverify fails on Windows because it's running a shell script--I'll adjust the platform restriction for that.

PR #788, opened 2026-06-05
fix: fix last release date in CHANGES

willkg:


willkg:
This is the last self-review for realz.

